PR #18, opened 2023-07-22
Add autofill for wordpress installation - makes every flag optional.

niristius:
I prototyped a way to make every app-installation-flag optional.
if at least one flag is missing the missing data will be generated from the current mStudio Token or other fallback-data.
Looking for a better way to get the local OS-User as fallback in src/commands/app/install/wordpress.tsx:104 without the require statement
Also: Im trying to refine the processs with WordPress as an exemplary process.
Creating the Processes for further Apps will just be a bit of copy&paste.


src/lib/app/flags.tsx
```diff
@@ -0,0 +1,286 @@
...
+    process.addInfo(
+      <Text>
+        Using generated random Admin Pass: <Value>{flags["admin-pass"]}</Value>
+      </Text>,
+    );
+  }
+
+  // Admin Firstname
+  if (necessaryFlags.includes("admin-firstname") && !flags["admin-firstname"]) {
+    flags["admin-firstname"] = ownUser.data.person!.firstName;
```
martin-helmich, line 234: We should test that `ownUser.data.person` is actually non-null.

src/lib/app/flags.tsx
```diff
@@ -0,0 +1,286 @@
...
+      <Text>
+        Using mStudio firstname as Admin firstname (
+        <Value>{flags["admin-firstname"]}</Value>)
+      </Text>,
+    );
+  }
+
+  // Admin Lastname
+  if (necessaryFlags.includes("admin-lastname") && !flags["admin-lastname"]) {
+    flags["admin-lastname"] = ownUser.data.person!.lastName;
```
martin-helmich, line 245: Same here

src/lib/app/flags.tsx
```diff
@@ -0,0 +1,286 @@
...
+    process.addInfo(
+      <Text>
+        Using mStudio firstname as Admin lastname (
```
martin-helmich, line 248:
```suggestion
        Using mStudio lastname as Admin lastname (
```

src/commands/app/install/contao.tsx
martin-helmich, line 1: Also, there's a new CI test that asserts that the `README.md` is up-to-date; can you run one final `yarn generate:readme`?

niristius: Ah, was searching for a way to generate it. Didn't think for it to be that easy👍🏼

martin-helmich (approved):

PR #29, opened 2023-08-01
Deduplicate code in app installer commands

martin-helmich:
- Add AppInstaller class to deduplicate installation commands
- Remove TODO
- Clean up stuff

### DIFF
--- a/src/commands/app/install/joomla.tsx
+++ b/src/commands/app/install/joomla.tsx
@@ -1,30 +1,16 @@
-import { normalizeToAppVersionUuid } from "../../../lib/app/versions.js";
-import {
-  autofillFlags,
-  provideSupportedFlags,
-  RelevantFlagInput,
-} from "../../../lib/app/flags.js";
-import { waitUntilAppIsInstalled } from "../../../lib/app/wait.js";
-import { MittwaldAPIV2 } from "@mittwald/api-client";
-import { withProjectId } from "../../../lib/project/flags.js";
 import { ExecRenderBaseCommand } from "../../../rendering/react/ExecRenderBaseCommand.js";
-import { makeProcessRenderer } from "../../../rendering/process/process_flags.js";
-import { Success } from "../../../rendering/react/components/Success.js";
 import React from "react";
-import AppAppVersion = MittwaldAPIV2.Components.Schemas.AppAppVersion;
-import { triggerAppInstallation } from "../../../lib/app/install.js";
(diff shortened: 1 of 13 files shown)


src/commands/app/install/joomla.tsx
```diff
@@ -1,111 +1,38 @@
...
+  ] as const,
+);
 
-  static description = `Creates new ${AppInstallation.appName} Installation.`;
-  static flags: RelevantFlagInput<typeof AppInstallation.appSupportedFlags> =
-    provideSupportedFlags(
-      AppInstallation.appSupportedFlags,
-      AppInstallation.appName,
-    );
+export default class InstallContao extends ExecRenderBaseCommand<
```
martin-helmich, line 24: This seems to be a consistent copy&paste issue

```suggestion
export default class InstallJoomla extends ExecRenderBaseCommand<
```

src/commands/app/install/node.tsx
```diff
@@ -1,100 +1,28 @@
...
-  { appInstallationId: string }
+import {
+  AppInstallationResult,
+  AppInstaller,
+} from "../../../lib/app/Installer.js";
+
+const installer = new AppInstaller(
+  "3e7f920b-a711-4d2f-9871-661e1b41a2f0",
+  "Node.js Project",
+  ["version", "site-title", "wait"] as const,
```
martin-helmich, line 11: I was just thinking: Since at least `"version"` and `"wait"` are _always_ there (and also `"site-title"`?): Should we extend the installer `get flags()` function to automatically include these flags, so that we don't have to explicitly define them for each installer?

src/commands/app/install/joomla.tsx
niristius: I completely disregarded this necessity.. its fixed now.

src/commands/app/install/node.tsx
niristius: There are a few pros and cons for this.

Pros:
- more unification
- code readability and less prone for errors

Cons:
- version would be even more forced for software that doesn't need it.
- - e.g. Node&PHP don't need a version flag since they are "abstract" Apps and only have and support Version Number 1.0.0
- we might have to bisect the flags.tsx all over again 

My know how does not suffice yet to gauge the full scope of necessary changes

martin-helmich: Yeah, I hadn't thought about the Node&PHP apps; that's a valid point. Let's leave the flags as they are for the moment, then.

PR #48, opened 2023-08-31
Improve behaviour for custom apps

### LINKED ISSUE #43: mw app list shows wrong PHP and Node.js version
If I show the list of available apps with ``mw app list``, it show "1.0.0" for PHP and Node.js, which isn't correct.

martin-helmich:
This PR improves the UX for installing custom (PHP + Node.js) applications. More precisely, this means:

- commands were moved from `mw app install {php,node}` to `mw app create {php,node}`, as it is more intuitive 
that the user does not "install an existing app", but instead "creates their own"
- "app list" and "app get" does not contain the version of the app template, because these are easily confused 
with the PHP or Node.js version that an app should use (fixes #43)

### DIFF
--- a/src/commands/app/create/node.tsx
+++ b/src/commands/app/create/node.tsx
@@ -5,24 +5,29 @@ import {
   AppInstaller,
 } from "../../../lib/app/Installer.js";
 
-const installer = new AppInstaller(
+export const nodeInstaller = new AppInstaller(
   "3e7f920b-a711-4d2f-9871-661e1b41a2f0",
-  "Node.js Project",
-  ["version", "site-title", "wait"] as const,
+  "custom Node.js",
+  ["site-title", "wait"] as const,
 );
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -80,15 +80,15 @@ USAGE
 
 <!-- commands -->
 * [`mw app copy INSTALLATION-ID`](#mw-app-copy-installation-id)
+* [`mw app create node`](#mw-app-create-node)
+* [`mw app create php`](#mw-app-create-php)
```
LukasFritzeDev, line 84: Is the renaming from `install` to `create` intentional? If so, I’d be interested in the reasoning, because this would introduce inconsistencies between these two apps and the other ones.

LukasFritzeDev (commented):
The rest looks good to me.

LukasFritzeDev (approved):

PR #144, opened 2023-12-19
Implement missing "cronjob" commands

martin-helmich:
- Add option to retrieve context values from terraform state
- Fix linter issues
- Regenerate README
- Make context provider named
- Adjust presentation of context values
- Add stricter typings for Terraform integration
- Remove superflous FlagIDType
- Extract app-installation determination into base class
- Make ListBaseCommand inherit ExtendedBaseCommand
- Move project id determination into class helper method
- Update README
- Add "cronjob create" + "delete" commands
- Add commands to trigger and cancel a cronjob execution
- Rebuild "cronjob get" command


src/rendering/react/components/CronJob/CronJobDetails.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <SingleResult
+      title="EXECUTION TARGET"
+      rows={{
+        URL: <Value>{dest.url}</Value>,
+      }}
+    />
+  );
+};
+
+const CronJobExecutionTargetCommand: FC<{ dest: CronjobCronjobCommand }> = ({
```
LukasFritzeDev, line 42: `dest`? wtf is `dest`? How do you get from `CronjobCronjobCommand` to `dest`? Can we name it `command` or `cmd` or something similar?

LukasFritzeDev (changes requested):

src/rendering/react/components/CronJob/CronJobDetails.tsx
martin-helmich: In the API spec, a cronjob has a "destination", which can be either an URL or a command -- which is where this naming comes from. I've renamed it to `command`, for this component.

README.md
```diff
@@ -1922,6 +1925,107 @@ FLAGS
...
+
+Create a new cron job
+
+```
+USAGE
+  $ mw cronjob create --description <value> --interval <value> [-i <value>] [-q] [--disable] [--email <value>]
+    [--url <value> | --command <value>] [--interpreter <value>]
+
+FLAGS
+  -i, --installation-id=<value>  ID or short ID of a installation; this flag is optional if a default installation is
```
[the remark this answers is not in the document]
martin-helmich: Renamed.

LukasFritzeDev (approved):

PR #218, opened 2024-02-13
Downgrade @mittwald/react-use-promise to ~2.1.2

martin-helmich:
This is to work around a known issue in @mittwald/react-use-promise
2.3 [1]

[1]: https://github.com/mittwald/react-use-promise/issues/157


martin-helmich:
Made superflous by https://github.com/mittwald/react-use-promise/pull/158

PR #220, opened 2024-02-15
Add DDEV integration

martin-helmich:
This PR adds a simple integration for [DDEV](https://ddev.readthedocs.io).

This integration consists of a few simple building blocks:

- The `mw ddev render-config` command may be used to create a project-/app-specific configuration file
- The `mw ddev init` command initializes a DDEV project in the local directory, whose configuration matches the 
specified mittwald app
- External to this repository, there is also the [mittwald/ddev addon](https://github.com/mittwald/ddev), which adds support for pushing+pulling code from+to a remote project


martin-helmich:
Blocked by https://github.com/mittwald/cli/pull/225, to enable automated testing.

src/commands/ddev/init.tsx
```diff
@@ -0,0 +1,150 @@
...
+    ]);
+  }
+
+  private async determineProjectName(r: ProcessRenderer): Promise<string> {
+    const { "project-name": projectName } = this.flags;
+    if (projectName) {
+      return projectName;
+    }
+
+    return await r.addInput("Enter the project name", false);
```
LukasFritzeDev, line 109: Improvement idea (maybe another PR): Use Dirname as default value. So if an empty name is given, the directory name is used.

src/lib/context_ddev.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Find the .ddev directory in the current working directory or any of its
+   * parent directories.
+   */
+  private async findDDEVConfigDir(): Promise<string | undefined> {
+    let currentDir = cwd();
+    while (currentDir !== "/") {
+      const ddevDir = path.join(currentDir, ".ddev", "config.yaml");
```
LukasFritzeDev, line 79: here it is
```ts
const ddevDir = path.join(currentDir, ".ddev", "config.yaml");
```

in `commands/ddev/init.tsx` it is:
```ts
const configFile = path.join(".ddev", "config.mittwald.yaml");
```

Can this be a problem, that in one case the `currentDir` is considered and in the other case it is not?

src/lib/ddev/config_builder.ts
```diff
@@ -0,0 +1,209 @@
...
+        inst.appVersion.desired,
+      );
+
+      const [major] = version.externalVersion.split(".");
+      return `drupal${major}`;
+    }
+
+    throw new Error(
+      "Automatic project type detection failed. Please specify the project type manually by setting the `--override-type` flag.",
+    );
```
LukasFritzeDev, line 95: Maybe use a `switch .. case` here

src/lib/ddev/config_builder.ts
```diff
@@ -0,0 +1,209 @@
...
+    systemSoftwareVersions: SystemSoftwareVersions,
+  ): string | undefined {
+    if (!("php" in systemSoftwareVersions)) {
+      return undefined;
+    }
+
+    const version = systemSoftwareVersions["php"];
+    const [major, minor] = version.split(".");
+
+    return `${major}.${minor}`;
```
LukasFritzeDev, line 129: Split on `.` and the join with `.`? 🤔 or is this to drop the patch level? 

Could be more intuitive if extracted `removePatchLevel(version)` 🤷

src/rendering/process/process_exec.ts
```diff
@@ -0,0 +1,39 @@
+import { spawn } from "child_process";
+import { ProcessRenderer } from "./process.js";
+import { ReactNode } from "react";
+
+export async function spawnInProcess(
+  p: ProcessRenderer,
```
LukasFritzeDev, line 6: ahhh … above (in the command files the Renderer is always named `r`, here it’s suddenly `p`

src/rendering/process/process_exec.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+  };
+
+  child.stdout.on("data", appendOutput);
+  child.stderr.on("data", appendOutput);
+  child.on("exit", (code) => {
+    if (code === 0) {
+      step.complete();
+    } else {
+      step.error(new Error(`rsync exited with code ${code}`));
```
LukasFritzeDev, line 34: Do you think only `rsync` might fail? 🤔 😜 

```suggestion
      step.error(new Error(`${cmd} exited with code ${code}`));
```

LukasFritzeDev (changes requested):

src/commands/ddev/init.tsx
martin-helmich: Partially unrelated, but that reminds me: I actually wanted to add a lookup to re-use the _existing_ project name, when the command is run in a directory where a DDEV project is already initialized...

src/rendering/process/process_exec.ts
martin-helmich: Yep, that's an oversight from refactoring. Good catch. :+1:

src/lib/ddev/config_builder.ts
martin-helmich: > or is this to drop the patch level?

Exactly this; refactoring incoming...

src/lib/ddev/config_builder.ts
martin-helmich: Incoming.

src/rendering/process/process_exec.ts
martin-helmich: Fixed.

src/lib/context_ddev.ts
martin-helmich: These are different use cases: The `ddev init` command is supposed to initialize a DDEV project in the current working directory of the user's shell; whereas the `DDEVContextProvider` is supposed to traverse the directory tree upwards to look for a ddev config dir _anywhere_ up the filesystem tree.

src/commands/ddev/init.tsx
martin-helmich: Regarding the default value: I wanted to mirror DDEV's behaviour here (which also prompts the user if no name is specified), as to not break anyone's expectations.

LukasFritzeDev: yes, I do agree with the behaviour to ask for a name if none is given. But what happens if I just hit enter when the input query is shown? Does it set an empty name? In this case the dirname could be a sensible fallback, if also showed as default.

LukasFritzeDev (approved):

src/commands/ddev/init.tsx
martin-helmich: Created #230 as follow-up

PR #226, opened 2024-02-22
Add "ssh-identity-file" flag and use it where appropriate

### LINKED ISSUE #57: Add IndentityFile support for `project ssh`
Using the native `ssh` command its possible to set an identity file using the option `-i`.

```
     -i identity_file
             Selects a file from which the identity (private key) for public key authentication is read.  You can also specify a public key file to use the corresponding private key that is loaded in ssh-agent(1) when the private key file is not present locally.  The default is
             ~/.ssh/id_rsa, ~/.ssh/id_ecdsa, ~/.ssh/id_ecdsa_sk, ~/.ssh/id_ed25519, ~/.ssh/id_ed25519_sk and ~/.ssh/id_dsa.  Identity files may also be specified on a per-host basis in the configuration file.  It is possible to have multiple -i options (and multiple identities specified
             in configuration files).  If no certificates have been explicitly specified by the CertificateFile directive, ssh will also try to load certificate information from the filename obtained by appending -cert.pub to identity filenames.
```

It would be great if the command `mw project ssh` would support a similar option. 

Since the identity file is user based it can be used for all projects. So another option would be to add a command like `mw login identity-file` to set the file globally, so it must not be provided every time the command `project ssh` is used.

martin-helmich:
Fixes #57


src/commands/database/redis/shell.tsx
```diff
@@ -8,12 +8,16 @@ import * as cp from "child_process";
...
+import { sshUsageDocumentation } from "../../../lib/ssh/doc.js";
 
 export class Shell extends ExecRenderBaseCommand<
   typeof Shell,
   Record<string, never>
 > {
   static summary = "Connect to a Redis database via the redis-cli";
+  static description =
+    "This command opens an interactive redis-cli shell to a Redis database.\n\n" +
+    sshUsageDocumentation;
```
LukasFritzeDev, line 20: should the new option be supported for this command as well? 

Currently the flag is not listed in the updated readme.

src/commands/database/mysql/shell.tsx
```diff
@@ -12,12 +12,16 @@ import {
...
+import { sshUsageDocumentation } from "../../../lib/ssh/doc.js";
 
 export class Shell extends ExecRenderBaseCommand<
   typeof Shell,
   Record<string, never>
 > {
   static summary = "Connect to a MySQL database via the MySQL shell";
+  static description =
+    "This command opens an interactive mysql shell to a MySQL database.\n\n" +
+    sshUsageDocumentation;
```
LukasFritzeDev, line 24: should the new option be supported for this command as well? 

Currently the flag is not listed in the updated readme.

martin-helmich: Good point! :+1:  Hang on... ⏳

src/commands/database/redis/shell.tsx
martin-helmich: Good catch. Fixed. 👌

src/commands/database/mysql/shell.tsx
martin-helmich: Also fixed 🙆‍♂️

LukasFritzeDev (approved):

PR #228, opened 2024-02-22
Refactor workflows for less duplication

martin-helmich:


gandie:
@martin-helmich i think this one is stale and can just vanish?

martin-helmich:
2,5 years-- yeah, let's close this.

PR #243, opened 2024-02-27
Add support for uploading app files + databases

martin-helmich:
This PR adds support for a `mw app upload` and a `mw database mysql import` command. Both of these are useful for offering a quick way to publish a development environment onto a server (most prominently, for using `ddev push`).

⚠️ It needs to be stressed that this is a development tool, and not a suitable substitute for a production-ready deployment solution.


README.md
```diff
@@ -1543,6 +1545,47 @@ FLAG DESCRIPTIONS
...
+
+ARGUMENTS
+  INSTALLATION-ID  ID or short ID of an app installation; this argument is optional if a default app installation is set
+                   in the context
+
+FLAGS
+  -q, --quiet             suppress process output and only display a machine-readable summary.
+      --delete            delete remote files that are not present locally
+      --dry-run           do not actually upload the app installation
+      --source=<value>    (required) source directory from which to upload the app installation
```
LukasFritzeDev, line 1564: Why is this a flag and not an argument? I assumed that flags _should_ be optional an required input is given as argument.

README.md
```diff
@@ -2375,6 +2418,65 @@ DESCRIPTION
...
+
+```
+USAGE
+  $ mw database mysql import DATABASE-ID -i <value> [-q] [-p <value>] [--temporary-user] [--ssh-user <value>] [--gzip]
+
+ARGUMENTS
+  DATABASE-ID  The ID or name of the database
+
+FLAGS
+  -i, --input=<value>           (required) the input file from which to read the dump ("-" for stdin)
```
LukasFritzeDev, line 2433: Why is this a flag and not an argument? I assumed that flags should be optional an required input is given as argument.

README.md
```diff
@@ -2375,6 +2418,65 @@ DESCRIPTION
   Get a MySQLDatabase.
 ```
 
+## `mw database mysql import DATABASE-ID`
+
+Imports a dump of a MySQL database
+
+```
+USAGE
+  $ mw database mysql import DATABASE-ID -i <value> [-q] [-p <value>] [--temporary-user] [--ssh-user <value>] [--gzip]
```
LukasFritzeDev, line 2427: Support for a [mysql options file](https://dev.mysql.com/doc/refman/8.0/en/option-files.html) would be awesome

`--defaults-extra-file="~/files/.my.cnf"`

This is only a feature idea and could also be a improvement for a follow up pull request.

README.md
LukasFritzeDev, line 1564: Would it be useful to set a default to the current directory if omitted?

README.md
```diff
@@ -1543,6 +1545,47 @@ FLAG DESCRIPTIONS
     scripts), you can use this flag to easily get the IDs of created resources for further processing.
 ```
 
+## `mw app upload [INSTALLATION-ID]`
```
LukasFritzeDev, line 1548: If I can _upload_ an app, it would be great to _download_ it as well 

This is only a feature idea and should be implemented in a follow up pull request.

src/commands/database/mysql/import.tsx
```diff
@@ -0,0 +1,132 @@
...
+      );
+
+    const { project } = connectionDetails;
+    const mysqlArgs = buildMySqlArgs(connectionDetails);
+
+    let cmd: RunCommand = { command: "mysql", args: mysqlArgs };
+    if (this.flags.gzip) {
+      const escapedArgs = shellEscape(mysqlArgs);
+      cmd = {
+        shell: `set -e -o pipefail > /dev/null ; gunzip | mysqldump ${escapedArgs}`,
```
LukasFritzeDev, line 68: Are you sure `mysqldump` is the correct command to **import** a sql file? 🤔

src/commands/database/mysql/import.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Text>
+        starting mysql via SSH on project <Value>{project.shortId}</Value>
+      </Text>,
+      () =>
+        executeViaSSH(
+          this.apiClient,
+          this.flags["ssh-user"],
+          { projectId: connectionDetails.project.id },
+          cmd,
+          null,
```
LukasFritzeDev, line 82: what's the meaning of this `null` function parameter? May we should refactor `executeViaSSH` to make the calls more readable

src/commands/database/mysql/import.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+}
+
+function buildMySqlArgs(d: {
+  hostname: string;
+  user: string;
+  password: string;
+  database: string;
+}): string[] {
+  return ["-h", d.hostname, "-u", d.user, "-p" + d.password, d.database];
```
LukasFritzeDev, line 131:
```suggestion
  return ["-h", d.hostname, "-u", d.user, `-p${d.password}`, d.database];
```

in my opinion, this would make it more optional, that this is on purpose and not a typo (`+` instead of `,`)

src/lib/database/mysql/temp_user.ts
```diff
@@ -0,0 +1,55 @@
...
+  user: DatabaseMySqlUser;
+  password: string;
+
+  cleanup(): Promise<void>;
+}
+
+/**
+ * Creates a temporary user for a database operation.
+ *
+ * Caution: The returned TemporaryUser object contains a "cleanup()" functions;
```
LukasFritzeDev, line 18:
```suggestion
 * Caution: The returned TemporaryUser object contains a "cleanup()" function;
```

src/lib/database/mysql/temp_user.ts
```diff
@@ -0,0 +1,55 @@
...
+ * callers of the createTemporaryUser function must make sure to call this
```
LukasFritzeDev, line 19: > „Functions should not have side effects“ \
> _https://www.youtube.com/watch?v=7EmboKQH8lM&t=5242s_

LukasFritzeDev (changes requested):

README.md
martin-helmich: That command already exists,[^1] added in #173 😉 

[^1]: https://github.com/mittwald/cli?tab=readme-ov-file#mw-app-download-installation-id

src/commands/database/mysql/import.tsx
martin-helmich: Ouch 😱

README.md
martin-helmich: I chose not to set a default on purpose, because this is a potentially very destructive command and I wanted the user to specify the upload source _explicitly_.

src/lib/database/mysql/temp_user.ts
martin-helmich: Thanks for the Uncle Bob Bomb... 😉  Refactored in d508120.

src/commands/database/mysql/import.tsx
martin-helmich: Fixed.

src/commands/database/mysql/import.tsx
martin-helmich: Refactored to use a parameter object with named fields for better readability.

src/commands/database/mysql/import.tsx
martin-helmich: Fixed.

README.md
martin-helmich: That's a valid point; however, this is built to mirror the flags of the (already existing) `app download` command -- to use an argument here, and a flag for `app download` would introduce further inconsistency. Ideally, both could (should) be changed to use arguments, but there are already [downstream integrations](https://github.com/mittwald/ddev) that rely on this interface.

In any case, I'd prefer to _not_ change this in this PR, but to give this more consideration in a follow-up issue.

README.md
martin-helmich: See https://github.com/mittwald/cli/pull/243#discussion_r1556058166

src/commands/database/mysql/dump.tsx
```diff
@@ -61,60 +50,44 @@ export class Dump extends ExecRenderBaseCommand<
     const databaseId = await withMySQLId(this.apiClient, this.flags, this.args);
     const p = makeProcessRenderer(this.flags, "Dumping a MySQL database");
 
-    const connectionDetails = await getConnectionDetailsWithPassword(
+    const name = await withConnectionDetails(
```
LukasFritzeDev, line 53: Is `name` the `userName` or `databaseName`/`dbName`? 😉

LukasFritzeDev, line 53: In most cases we would call a such a function `run...`, in this case `runWithConnectionDetails`. I can also live with the current name, I would leave the decision up to you.

martin-helmich: Fixed with c2b21c04ef61497d6105ef83190dc26d299b004b

martin-helmich: Fixed with bcef63d3954d3995131d6dc1b1cc3cbb6b66a802.

README.md
martin-helmich: Created #393 as a follow-up issue.

LukasFritzeDev:
Look good

LukasFritzeDev (approved):

README.md
martin-helmich: Created #394 as follow-up

PR #395, opened 2024-04-09
Draft: add app upgrade command 

niristius:
Add app upgrade command to be able to trigger app upgrades.


src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }),
+    "target-version": Flags.string({
+      description: "target version to upgrade target app to",
+      required: true,
+      default: "latest",
+    }),
+  };
+
+  protected async exec(): Promise<boolean> {
+    let appInstallationId: string = checkInstallationIdInput(
```
martin-helmich, line 29: There are predefined methods for getting the app installation id from args/flags/context; please just use the `withAppInstallationId` function, and drop the flag and the `checkInstallationIdInput` function.

PR #399, opened 2024-04-10
Feat: Add app upgrade

niristius:
This feature is not yet done.
There are still some problems with choosing specific target versions for an upgrade.
Also prompts, user interaction aswell as general code efficiency and hygiene needs some more work.
Any feedback is already welcome though!


src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,170 @@
...
+  };
+  static flags = {
+    "target-version": Flags.string({
+      description: "target version to upgrade target app to",
+      required: true,
+      default: "latest",
+    }),
+    "show-candidates": Flags.boolean({
+      description: "list version candidates the given app can be upgraded to",
+    }),
```
martin-helmich, line 35: Listing potential target versions and actually performing an update are quite different things; maybe extract this functionality into a different command? Like `mw app list-upgrade-candidates`?

EDIT: That command would also make sense to be implemented as a `ListBaseCommand` subclass, because I could imagine a user wanting JSON/YAML output for that command (which would not be easily possible with the current implementation).

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,170 @@
...
+      UpgradeApp,
+      this.flags,
+      this.args,
+      this.config,
+    );
+
+    const currentAppInstallationResponse =
+      await this.apiClient.app.getAppinstallation({
+        appInstallationId: appInstallationId,
+      });
```
martin-helmich, line 60: `assertStatus` to assert the expected return status code; this should also make the subsequent type casts unnecessary, because the compiler should be able to figure out the data types on its own.

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,170 @@
...
+      process.error("Target App version could not be determined properly.");
+      return;
+    }
+
+    if (!this.flags.force) {
+      const confirmed = await process.addConfirmation(
+        <Text>
+          Confirm Upgrading {currentApp.name}{" "}
+          {currentAppVersion.externalVersion} (description here) to Version{" "}
+          {targetAppVersion.externalVersion}
```
martin-helmich, line 138:
```suggestion
          Confirm upgrading {currentApp.name}{" "}
          {currentAppVersion.externalVersion} (description here) to version{" "}
          {targetAppVersion.externalVersion}
```

https://www.grammarly.com/blog/capitalization-rules/ 😉

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,170 @@
...
+        process.addInfo(<Text>Upgrade will not be triggered.</Text>);
+        process.complete(<></>);
+
+        ux.exit(1);
+      }
+    } else {
+      process.addInfo(
+        <Text>
+          Commencing upgrade of {currentApp.name}
+          {currentAppVersion.externalVersion} (description here) to Version
```
martin-helmich, line 151:
```suggestion
          {currentAppVersion.externalVersion} (description here) to version
```

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    this.apiClient.app.patchAppinstallation({
+      appInstallationId,
+      data: { appVersionId: targetAppVersion.id },
+    });
+
+    process.complete(
+      <Success>The App Upgrade has been started. Buckle up! 🚀</Success>,
+    );
```
martin-helmich, line 164: Maybe also add a `--wait` flag (like for the `install [appname]` commands), in case someone wants to use this command synchronously?

src/lib/app/versions.tsx
```diff
@@ -70,6 +69,46 @@ export async function getLatestAvailableAppVersionForApp(
...
+    appId,
+    baseAppVersionId,
+  });
+  assertStatus(versions, 200);
+  if (versions.data.length === 0) {
+    return undefined;
+  }
+  let latestVersion = "0.0.0";
+  for (const version of versions.data) {
+    if (gt(version.internalVersion, latestVersion)) {
```
martin-helmich, line 87: If we're using semver functionalities anyway: Should we just offer the option for the user to specify a semver selector? Like...

```
$ mw app upgrade <app-id> --target-version="^12.4.0"
$ mw app upgrade <app-id> --target-version="*"
```

Consequently, the `install` commands should then support this, as well.

Could also be implemented as a follow-up PR.

src/commands/app/upgrade.tsx
martin-helmich: Also, what's up with `description here`?

martin-helmich (changes requested):

src/commands/app/upgrade.tsx
niristius: Valid point - This will be implemented.
I would argue though, that since most of the same data is required for the upgrade process anyway, so giving the user the option to check the candidates while already preparing the upgrade command might make for a smoother workflow - Albeit this means that either the flag or the extra command would be redundant. I don't see the harm though except some degree of bloating the cli?

martin-helmich: Or, alternatively, while thinking about it some more: If no `--target-version` was specified, give the user the option to select the desired target version interactively (if used within a TTY context) from the candidate list; this could be done using the `ProcessRenderer`s `addSelect` method:

https://github.com/mittwald/cli/blob/d8076f7877613b86aaebabeb8bf55654003245fd/src/rendering/process/process.tsx#L116-L119

src/commands/app/upgrade.tsx
niristius: mentioned typos have been corrected, and then some.

src/lib/app/versions.tsx
niristius: I like this.
I'll add it as a todo for this PR, but we'll see if we don't just merge somewhere in between already and make a new PR with less impactful stuff

src/commands/app/upgrade.tsx
niristius: it's on the todo list, just wanted to get basic functionalities out of the way first and then look into how the waiting at that point actually works before I just copy&paste functionalities.

src/commands/app/upgrade.tsx
niristius: fixed in 25a4535f81bd2aa692f974e7819dc733581dab6c

src/commands/app/upgrade.tsx
niristius: I'll have to keep an eye out for correct assertion, and will do that.
For this explicit instance the whole api call was put into a wrapper function with correct assertion in 2fe9d63108be465704faa55aae270c26dfb4f3ce

niristius:
Still Todo

- [ ] fix up the "list-appupgrade-candidates" command to actually make it work
- [ ] add app installation description to user dialogue where it make sense
- [ ] make appinstallation selection interactive if none is given
- [ ] make upgrade target version selection interactive if none given
- [ ] add "--wait" flag 
- [ ] target version semver functionalities

niristius:
This PR is ready to be merged after having gotten a second opinion on user dialogue and general functionality.
I merged every commit that was pushed into master since this branch was created and generated the readme but the readme check still fails, what am I missing?

About the two open todos:

Making the Project or App as an interactive selection would need a bit of work on the "withProjectId"/"withAppInstallationId" functions as to not bloat or overcomplicate things. I dont see it as a necessary part of this PR's feature.

I will tackle target version semver functionalities in a follow up PR at a later date.

README.md
```diff
@@ -1932,7 +1996,7 @@ EXAMPLES
   $ mw autocomplete --refresh-cache
 ```
 
-_See code: [@oclif/plugin-autocomplete](https://github.com/oclif/plugin-autocomplete/blob/v3.0.16/src/commands/autocomplete/index.ts)_
+_See code: [@oclif/plugin-autocomplete](https://github.com/oclif/plugin-autocomplete/blob/v3.0.13/src/commands/autocomplete/index.ts)_
```
martin-helmich, line 1999: Looks like you somehow DOWNgraded some of the dependencies. Since the `yarn.lock` is unmodified → did you perhaps forget to install the latest dependencies before generating the README?

src/commands/app/list-upgrade-candidates.ts
```diff
@@ -0,0 +1,68 @@
...
+type Response = Awaited<
+  ReturnType<MittwaldAPIV2Client["app"]["listAppversions"]>
+>;
+
+export default class List extends ListBaseCommand<
+  typeof List,
+  ResponseItem,
+  Response
+> {
+  static description = "List update candidates for an appInstallation.";
```
martin-helmich, line 24: Wording consistency: "upgrade" vs. "update"

```suggestion
  static description = "List upgrade candidates for an app installation.";
```

src/commands/app/list-upgrade-candidates.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const currentAppInstallation = await getAppInstallationFromUuid(
+      this.apiClient,
+      appInstallationId,
+    );
+
+    return await this.apiClient.app.listUpdateCandidatesForAppversion({
+      appId: currentAppInstallation.appId,
+      baseAppVersionId: (
+        currentAppInstallation.appVersion as { current: string }
```
martin-helmich, line 49: Why is this typecast necessary?

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,230 @@
...
+  currentApp: AppApp,
+  currentAppVersion: AppAppVersion,
+): Promise<AppAppVersion | undefined> {
+  const targetAppVersionString = await process.addSelect(
+    `Please select target upgrade for your ${currentApp.name} ${currentAppVersion.externalVersion} from one of the following`,
+    [
+      {
+        value: "latest",
+        label: "latest",
+      },
```
martin-helmich, line 208: Do we actually need this special "latest" option? The actual latest version should be the first in the candidate list anyway, shouldn't it?

src/lib/app/uuid.ts
```diff
@@ -50,6 +50,29 @@ export async function getAppVersionFromUuid(
...
+ * @param apiClient
+ * @param appInstallationId
+ */
+export async function getAppInstallationFromUuid(
+  apiClient: MittwaldAPIV2Client,
+  appInstallationId: string,
+): Promise<AppAppInstallation> {
+  if (!isUuid(appInstallationId)) {
+    throw new Error("Given UUID not valid.");
+  }
```
martin-helmich, line 65: tbh., I'd just rely on the API's response code here, instead of manually validating the input query. If it's not a valid installation ID, the API will tell you soon enough. 🤷

src/lib/app/versions.tsx
```diff
@@ -34,13 +33,13 @@ export async function normalizeToAppVersionUuid(
       `${await getAppNameFromUuid(
         apiClient,
         appUuid,
-      )} version ${version} does not seem to exist for the mStudio.`,
+      )} Version ${version} does not seem to exist for the mStudio.`,
```
martin-helmich, line 36: Capitalization was correct before.

```suggestion
      )} version ${version} does not seem to exist for the mStudio.`,
```

src/lib/app/versions.tsx
```diff
@@ -34,13 +33,13 @@ export async function normalizeToAppVersionUuid(
...
     );
   }
 
   process.addInfo(
     <Text>
-      installing version: <Value>{appVersion.externalVersion}</Value>
+      installing Version: <Value>{appVersion.externalVersion}</Value>
```
martin-helmich, line 42: Capitalization was correct before.

```suggestion
      installing version: <Value>{appVersion.externalVersion}</Value>
```

src/lib/app/versions.tsx
```diff
@@ -70,7 +69,64 @@ export async function getLatestAvailableAppVersionForApp(
...
+  }
+  let latestVersion = "0.0.0";
+  for (const version of versions.data) {
+    if (gt(version.internalVersion, latestVersion)) {
+      latestVersion = version.internalVersion;
+    }
+  }
+  return versions.data.find(
+    (item: AppVersion) => item.internalVersion === latestVersion,
+  );
```
martin-helmich, line 111: Can be simplified; just store the entire version object as `latestVersion` and save yourself the lookup at the end:

```suggestion
  let latestVersion: AppVersion|undefined;
  for (const version of versions.data) {
    if (gt(version.internalVersion, latestVersion?.internalVersion ?? "0.0.0")) {
      latestVersion = version;
    }
  }
  return latestVersion;
```

src/lib/app/wait.tsx
```diff
@@ -29,3 +29,29 @@ export async function waitUntilAppIsInstalled(
 
   stepWaiting.complete();
 }
+
+export async function waitUntilAppIsUpgraded(
```
martin-helmich, line 33: Apart from the label text, this function is identical to `waitUntilAppIsInstalled`. This is a lot of duplication and should be simplified.

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,230 @@
...
+import { MittwaldAPIV2, MittwaldAPIV2Client } from "@mittwald/api-client";
+import { waitUntilAppIsUpgraded } from "../../lib/app/wait.js";
+import { assertStatus } from "@mittwald/api-client-commons";
+
+type AppApp = MittwaldAPIV2.Components.Schemas.AppApp;
+type AppAppInstallation = MittwaldAPIV2.Components.Schemas.AppAppInstallation;
+type AppAppVersion = MittwaldAPIV2.Components.Schemas.AppAppVersion;
+
+export class UpgradeApp extends ExecRenderBaseCommand<typeof UpgradeApp, void> {
+  static description = "Upgrade target appinstallation to target version";
```
martin-helmich, line 36:
```suggestion
  static description = "Upgrade app installation to target version";
```

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,230 @@
...
+  static args = {
+    ...appInstallationArgs,
+  };
+  static flags = {
+    "target-version": Flags.string({
+      description: "Target version to upgrade target app to.",
+    }),
```
martin-helmich, line 43: This this flag is neither required nor has a default value, the documentation should explain what happens when the flag is omitted.

```suggestion
    "target-version": Flags.string({
      description: "target version to upgrade app to; if omitted, target version will be prompted interactively",
    }),
```

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,230 @@
...
+        UpgradeApp,
+        this.flags,
+        this.args,
+        this.config,
+      ),
+      currentAppInstallation: AppAppInstallation =
+        await getAppInstallationFromUuid(this.apiClient, appInstallationId),
+      currentApp: AppApp = await getAppFromUuid(
+        this.apiClient,
+        currentAppInstallation.appId as string,
```
martin-helmich, line 72: Why is the `as string` typecast necessary?

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,230 @@
...
+      ),
+      currentAppVersion: AppAppVersion = await getAppVersionFromUuid(
+        this.apiClient,
+        currentApp.id,
+        (currentAppInstallation.appVersion as { current: string }).current,
```
martin-helmich, line 77: Why is the `as { current: string }` typecast necessary?

martin-helmich (changes requested):

README.md
niristius: I was not aware of this, fixed with 5b1b3a4230e7dd732015d94c52e86267b50156bb

src/commands/app/list-upgrade-candidates.ts
niristius: It isn't that is a inadequate quickfix I forgot to properly handle because I didn't know how at that time.
Follow up question to that one though: The problem, wich I didnt understand at that time, was just that `baseAppVersionId` expects a string, and `currentAppInstallation.appVersion.current` might be `undefined`. So the solutaion is a basic check if it is undefined.

Now to my question: Usually I would use `process.error()` and let the process handle the error.
In the `ListBaseCommand` we dont have a process though. Do I just `throw.error()`? (90d3c58abce4f3cc34817428761ae64beeac3e6a)

martin-helmich: > So the solution is a basic check if it is undefined.

Please note that the typecast does not _actually_ check if `.current` is undefined; you just tell the compiler that it's not (and it will fail at runtime, if it turns out to be undefined, anyway).

You could either assert that `.current` is indeed a string (`typeof currentAppInstallation.appVersion.current === "string"` should be sufficient for the compiler) and throw an error if not (the error will be handled properly by the framework itself);

Alternatively, you might also check against `.desired`, which IRRC is not-nullable -- which might just work better anyway, in case a user decides to invoke `list-upgrade-candidates` on an app installation for which an upgrade is currently in progress.

src/commands/app/upgrade.tsx
niristius: You aren't wrong - I just thought that giving the latest option would be more in unison with the option to declare the `--target-version` as `latest`, this might be personal preference though.

Also, the list returned by `apiClient.app.listUpdateCandidatesForAppversion()` is not ordered by semver.

- Ordering it during runtime for the interactive selection is not a big task but I'd argue its the wrong layer to do it on.
- Ordering it for the `list-upgrade-candidates`command might be though since this is a direct digest of the api-client-response. 
- Ordering it in the api-client before returning it might be a weird way to handle it since the api-client should contain as little logical digestion as possible.

I would argue this should already be returned by the api as a semver sorted list.
And as soon as the list os actually ordered, the `latest` entry might as well be purged.

src/lib/app/uuid.ts
niristius: My train of thought was that this function has no idea where the input comes from, it just gets passed a `string`.
If the given parameters came from an api-response or not should not be relevant for this function explicitly at the time of that check - or is that too petty?

src/lib/app/wait.tsx
niristius: Oh yeah, you're completely right.
I propose using one "waitUntilAppStateHasNormalized"-function with one additional parameter wich is the label.
ab1ab4b251d7b27c441b34889709bbc39299a3f6

src/commands/app/upgrade.tsx
niristius: As above, this is a workaround since I didnt understand that its just a missing check if it even is undefined.
Fixed with a check and exit with error code 1 if its undefined.

src/commands/app/upgrade.tsx
niristius: It's not. my personal skill issue, I assume. its gone.

src/commands/app/upgrade.tsx
niristius: sort implemented

src/lib/app/uuid.ts
niristius: Oh, I've only just figured out how you meant that. Feed the string into the getAppinstallation request and see if the client/api can work with the string. Consider it done!

src/commands/app/upgrade.tsx
```diff
@@ -0,0 +1,238 @@
...
+      currentAppVersion.id,
+    );
+  } else {
+    return await getAvailableTargetAppVersionFromExternalVersion(
+      apiClient,
+      currentApp.id,
+      currentAppVersion.id,
+      targetAppVersionString,
+    );
+  }
```
martin-helmich, line 237: If `latest` is not a selectable option anymore, this condition shouldn't be necessary anymore, either.

```suggestion
  return await getAvailableTargetAppVersionFromExternalVersion(
    apiClient,
    currentApp.id,
    currentAppVersion.id,
    targetAppVersionString,
  );
```

niristius: latest was removed from the interactive selection only.
You can still trigger an upgrade to `latest` using the  `--target-version` flag (as of now).

niristius: Should I remove it here aswell? I thought it would be a helpful option especially for custom update scripts.

niristius: @martin-helmich delete or keep the option to use "latest" for the "target-version"-flag - what's your opinion?
As soon as that is decided eventual changes take about a minute - after that, this feature can be merged.

martin-helmich: The option to select the latest version via flag is completely valid. But... this function appears to *only* cover interactive selection (the first statement is an unconditional `await process.addSelect`). If `latest` is not an option for interactive selection, I see no way how `targetAppVersionString` could have `"latest"` as value.

martin-helmich (approved):
LGTM now. :+1:   (only watch out for the merge conflict)

src/commands/app/upgrade.tsx
niristius: I was not paying sufficient attention to the line numbers. 
You are completely right and I need to be more careful.

Fixed it immediately.

martin-helmich (approved):

PR #422, opened 2024-04-22
feat: add mail address update

### LINKED ISSUE #232: Add command to edit mail address
I just created some mail addresses with mailboxes using `mw mail address create --random-password --enable-spam-protection --quiet --address ...`. After this I wanted to ad a forward address in addition to the mailbox. This is currently not possible  with the CLI.

* Add command to edit an existing mail address:
  * add forwardAddresses
  * de-/activate spam protection
  * ...
* add option to `mw mail address create` to force a mailbox even if `--forward-to` option is set

tillwestpfahl:
resolves #232.


src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,267 @@
...
+        },
+      });
+
+      assertStatus(response, 204);
+    });
+    await process.complete(
+      <Success>
+        The quota of your mail address was successfully updated.
+      </Success>,
+    );
```
martin-helmich, line 210: `process.complete` is not designed to be invoked multiple times. I'd recommend calling it at the end of `exec()` when all updates have been completed.

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,267 @@
...
+    "catch-all": Flags.boolean({
+      description: "make this a catch-all mail address",
+    }),
+    quota: Flags.integer({
+      description: "mailbox quota in mebibytes",
+    }),
+    password: Flags.string({
+      summary: "mailbox password",
+      description:
+        "This is the password that should be used for the mailbox; if omitted, the command will prompt interactively for a password.\n\nCAUTION: providing this flag may log your password in your shell history!",
```
martin-helmich, line 52: > if omitted, the command will prompt interactively for a password.

In the case of the `update` command, this is not correct. As it's built now (correctly), the password just remains unchanged when no `--password` flag was passed to the command.

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,267 @@
...
+    A mail address is either associated with a mailbox, or forwards to another address.
+    
+    To set forwarding addresses, use the --forward-to flag.
+    
+    Use the --catch-all flag to make the mailbox a catch-all mailbox.
+  
+    When running this command with --generated-password the output will be the newly generated and set password.`;
+  static args = { ...mailAddressArgs };
+  static flags = {
+    ...projectFlags,
```
martin-helmich, line 37: This command should not require a `--project-id` flag, as it's always referring to an existing mail address that's already associated with a project.

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,267 @@
...
+  protected async setQuota(
+    mailAddressId: string,
+    quota: number,
+    process: ProcessRenderer,
+  ): Promise<void> {
+    await process.runStep("set the quota of a mail address", async () => {
+      const response = await this.apiClient.mail.updateMailAddressQuota({
+        mailAddressId: mailAddressId,
+        data: {
+          quotaInBytes: quota,
```
martin-helmich, line 200: This says `quotaInBytes`, but according to the command documentation, the `--quota` flag is "mailbox quota in mebibytes"

src/lib/mail/flags.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (isUuid(id)) {
+      return id;
+    }
+
+    const response = await apiClient.mail.listMailAddresses({
+      projectId,
+    });
+    assertStatus(response, 200);
+
+    const address = response.data.find((address) => address.id === id);
```
martin-helmich, line 20: This is unnecessary, since you can just look-up mail addresses by their ID directly. This would only make sense if you would want to use the address string in place of an ID (like with domains, where we allow using the domain name instead of the domain ID).

martin-helmich (changes requested):

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,267 @@
...
+          flags: [
+            "catch-all",
+            "enable-spam-protection",
+            "quota",
+            "password",
+            "random-password",
+          ],
+        },
+      ],
+    }),
```
martin-helmich, line 76: Also, `create` and `update` share most of their flags. Perhaps it would make sense to extract these into a common `mailaddressFlags` object that can be spreaded into the flag definitions of both commands (see `mysql` commands, for exampe).

src/commands/mail/address/update.tsx
tillwestpfahl: Which command documentation? If I understand the command handler code correctly, this quotaInBytes is correct.

src/lib/mail/flags.ts
martin-helmich: ACK, but shouldn't it be `address.address === id`, then?

src/commands/mail/address/update.tsx
martin-helmich: I mean the documentation string of the `--quota` flag. Declared in the `flags` property of this class.

tillwestpfahl: Ah, yes i changed it to bytes in the documentation of both the create and update commands.

src/commands/mail/address/update.tsx
tillwestpfahl: I extracted `address` & `"catch-all"`. There are other shared flags, but they differ slightly between update and create.

src/commands/mail/address/update.tsx
tillwestpfahl: Oooooh, I completely misunderstood that. Yeah, that makes sense. I changed both descriptions back to mebibytes and added mebibyte to byte conversion to `update`.

README.md
```diff
@@ -3691,6 +3692,78 @@ FLAG DESCRIPTIONS
...
+    [--random-password] [--forward-to <value>]
+
+ARGUMENTS
+  MAILADDRESS-ID  ID or mail address of a mailaddress
+
+FLAGS
+  -a, --address=<value>        (required) mail address
+  -q, --quiet                  suppress process output and only display a machine-readable summary.
+      --catch-all              make this a catch-all mail address
+      --forward-to=<value>...  forward mail to another address
```
martin-helmich, line 3711: Documentation question: Are the `--forward-to` addresses additive, or do these values replace already existing forward addresses?

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,236 @@
...
+  protected async setCatchAll(
+    mailAddressId: string,
+    catchAll: boolean,
+    process: ProcessRenderer,
+  ): Promise<void> {
+    await process.runStep("set the catchall of a mail address", async () => {
+      const response = await this.apiClient.mail.updateMailAddressCatchAll({
+        mailAddressId: mailAddressId,
+        data: {
+          active: catchAll,
```
martin-helmich, line 113: `active`? Wrong field?

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,236 @@
...
+      this.flags,
+      this.args,
+      this.config,
+    );
+
+    if (this.flags.address) {
+      this.setAddress(mailAddressId, this.flags.address, process);
+    }
+
+    this.setCatchAll(mailAddressId, this.flags["catch-all"], process);
```
martin-helmich, line 192: This is dangerous; all other flags follow the semantic "leave unchanged, if the flag is not set" -- however, when you edit a catch-all address without `--catch-all`, this command would unset the catch-all setting.

This is easier on the `create` command (`--catch-all` means catch-all, and no `--catch-all` means no catch-all), whereas on `update` it might make more sense to actually support `--catch-all` and `--no-catch-all`.

README.md
tillwestpfahl: They completely replace the existing array of forward-addresses.

src/commands/mail/address/update.tsx
tillwestpfahl: Nope, thats just how bool is called in the openApi 🤷‍♂️

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,254 @@
...
+  static flags = {
+    ...processFlags,
+    address: Flags.string({
+      char: "a",
+      summary: "mail address",
+    }),
+    "catch-all": Flags.boolean({
+      description: "make this a catch-all mail address",
+    }),
+    "no-catch-all": Flags.boolean({
```
martin-helmich, line 45: `Flags.boolean` allows `allowNo: true`, which makes the `--no-${flagname}` implicit (in which case you'd need to differentiate between `false` and `undefined`)

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,254 @@
...
+      description: "make this not a catch-all mail address",
+    }),
```
martin-helmich, line 47:
```suggestion
    "catch-all": Flags.boolean({
      description: "make this a catch-all mail address",
      allowNo: true,
    }),
```

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,254 @@
...
+      this.setAddress(mailAddressId, this.flags.address, process);
+    }
+
+    if (this.flags["catch-all"]) {
+      this.setCatchAll(mailAddressId, true, process);
+    }
+
+    if (this.flags["no-catch-all"]) {
+      this.setCatchAll(mailAddressId, false, process);
+    }
```
martin-helmich, line 210:
```suggestion
    if (this.flags["catch-all"] !== undefined) {
      this.setCatchAll(mailAddressId, this.flags["catch-all"], process);
    }
```

src/commands/mail/address/update.tsx
tillwestpfahl: > make this a catch-all mail address; omit to leave unchanged
What if I want to change from a catch-all to a regular mail address? Does that still work?

src/commands/mail/address/update.tsx
```diff
@@ -0,0 +1,249 @@
...
+    const mailAddressId = await withMailAddressId(
+      this.apiClient,
+      Update,
+      this.flags,
+      this.args,
+      this.config,
+    );
+
+    if (this.flags.address) {
+      this.setAddress(mailAddressId, this.flags.address, process);
```
martin-helmich, line 200: Oh, I just noticed this right now! 😱   All these `set[...]` functions are `async`, but none of them are `awaited` anywhere.

I'm wondering why the linter didn't catch this... 🤔

tillwestpfahl: I saw a few cases of "missing await" warnings (of Phpstorm) in multiple files, but ignored them, because I am not very knowledgable at that topic. I fixed it know in all the mail and domain files.

martin-helmich (approved):

PR #438, opened 2024-04-29
Interactively prompt for DDEV project type if undiscoverable

### LINKED ISSUE #434: DDEV: Error: Automatic project type detection failed.
I have a App of type `Custom PHP app` with a manually installed WordPress.

When trying to init an DDEV project using `mw ddev init` I get this Error:
```
Error: Automatic project type detection failed.
```
Bildschirmfoto 2024-04-26 um 13 45 28

I think in this case we have multiple options to handle this more gracefully:
1. prompt for the project type (as it’s done already for project name and database)
1. use `--project-type=ddev` as a default for DDEV

martin-helmich:
This PR adds an interactive prompt when the DDEV project type cannot be determined automatically.

Fixes #434

### DIFF
--- a/src/lib/ddev/init_projecttype.tsx
+++ b/src/lib/ddev/init_projecttype.tsx
@@ -1,0 +1,145 @@
+import type { MittwaldAPIV2 } from "@mittwald/api-client";
+import { MittwaldAPIV2Client, assertStatus } from "@mittwald/api-client";
+import { typo3Installer } from "../../commands/app/install/typo3.js";
+import { wordpressInstaller } from "../../commands/app/install/wordpress.js";
+import { shopware6Installer } from "../../commands/app/install/shopware6.js";
+import { drupalInstaller } from "../../commands/app/install/drupal.js";
+import { ProcessRenderer } from "../../rendering/process/process.js";
+import { Value } from "../../rendering/react/components/Value.js";
+import { Text } from "ink";
+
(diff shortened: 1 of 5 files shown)


src/lib/ddev/init_projecttype.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return "shopware6";
+    case drupalInstaller.appId: {
+      const version = await getAppVersion(
+        client,
+        inst.appId,
+        inst.appVersion.desired,
+      );
+
+      const [major] = version.externalVersion.split(".");
+      return `drupal${major}`;
```
LukasFritzeDev, line 123: Does this really work?

Available ddev project types (defined in line 22–24): `"drupal6", "drupal7", "drupal",`.  In the mStudio GUI it’s currently possible to install version `10.2.5`. This would result in `project-type=drupal10`, wouldn’t it?

martin-helmich: Ouch, good catch. 😖   I have to admit that I was completely oblivious as to which Drupal versions actually exist out there. Should've done my homework better... 😕 

Hang on...

LukasFritzeDev (approved):

src/lib/ddev/init_projecttype.tsx
martin-helmich: Fixed in 8f96ac6

LukasFritzeDev (approved):

PR #465, opened 2024-05-08
Feat: add cronjob update

niristius:


niristius:
@martin-helmich I generalized the flags now to reduce duplication. Im struggling with the fact that I dont know how to make certain flags required for the create process while all flags are optional for the update process. Can you help me out? Once this works I will implement this approach for all my other PRs

src/commands/cronjob/create.tsx
```diff
@@ -97,18 +69,50 @@ export class Create extends ExecRenderBaseCommand<typeof Create, Result> {
...
+      if (interpreter == "bash") {
+        destinationInterpreter = "/bin/bash";
+      } else {
+        destinationInterpreter = "/usr/bin/php";
+      }
+
+      destination = {
+        interpreter: destinationInterpreter,
+        path: command as string,
+      };
```
martin-helmich, line 91: This is largely duplicated with the same part from the `update` command. Maybe extract this into a common function (something like `function buildCronjobDestination(url: string|undefined, command: string|undefined, interpreter: string|undefined): CronjobCreationData["destination"]`)?

src/commands/cronjob/create.tsx
```diff
@@ -97,18 +69,50 @@ export class Create extends ExecRenderBaseCommand<typeof Create, Result> {
...
+    } else if (!url && !command) {
+      throw new Error("either `--url` or `--command` must be specified");
+    } else if (command && !interpreter) {
+      throw new Error("--interpreter flag is required when command is given");
+    } else if (interpreter && !command) {
+      throw new Error("--command flag is required when interpreter is given");
+    } else {
+      throw new Error("something went wrong trying to digest destination.");
```
martin-helmich, line 99: These constraints _might_ also be configurable using the flag definitions.

martin-helmich (changes requested):

src/commands/cronjob/create.tsx
```diff
@@ -4,65 +4,49 @@ import {
...
-      default: undefined,
-    }),
-    interval: Flags.string({
-      required: true,
-      summary: "Interval of the cron job, in standard UNIX cron syntax",
-      default: undefined,
+    description: cronjobFlagDefinitions.description({ required: true }),
+    interval: cronjobFlagDefinitions.interval({ required: true }),
+    email: cronjobFlagDefinitions.email(),
+    timeout: cronjobFlagDefinitions.timeout({ required: true }),
```
martin-helmich, line 33: Changing the `--timeout` flag from a "duration" flag to a "number" flag is a UX downgrade, IMO. Before, you could specify values like `1h`, now you need to know how many seconds an hour has. Also, I think I would prefer having a (sensible) default value here, instead of this field being required.

src/commands/cronjob/update.tsx
```diff
@@ -0,0 +1,133 @@
...
+    }),
+  };
+
+  protected async exec(): Promise<void> {
+    const process = makeProcessRenderer(this.flags, "Updating cron job");
+    const { "cronjob-id": cronjobId } = this.args;
+    const currentCronjob = await this.apiClient.cronjob.getCronjob({
+      cronjobId,
+    });
+    assertSuccess(currentCronjob);
```
martin-helmich, line 66: I can't see `currentCronjob` being used anywhere. Is this a leftover from a previous iteration?

src/lib/resources/cronjob/destination.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    destination = {
+      interpreter: destinationInterpreter,
+      path: command as string,
+    };
+  } else {
+    return undefined;
+  }
+  return destination;
```
martin-helmich, line 31: Since this is a single function, you can drop the result variable, and just `return` directly. This also reduces the conditional nesting in this function quite a bit:

```suggestion
  if (url) {
    return { url };
  } 
  
  if (command && interpreter) {
    let destinationInterpreter;
    if (interpreter == "bash") {
      destinationInterpreter = "/bin/bash";
    } else {
      destinationInterpreter = "/usr/bin/php";
    }

    return {
      interpreter: destinationInterpreter,
      path: command as string,
    };
  }
  
  return undefined;
```

Could be reduced even further, if the whole `destinationInterpreter` logic would be extracted into a separate function (say, `mapInterpreterToFullPath`):

```suggestion
  if (url) {
    return { url };
  }
  
  if (command && interpreter) {
    return {
      interpreter: mapInterpreterToFullPath(interpreter),
      path: command as string,
    };
  }

  return undefined;
```

martin-helmich (changes requested):

src/commands/cronjob/create.tsx
niristius: I couldnt figure out how to implement this with the centralized flags.
`timeout: Flags.custom<Duration.relativeFlag>()` seems nonsensical.
Seem to be "too custom for custom flags"
So I rolled that flag back to how it worked before since you're abeolutely right, the new implementation was a downgrade in every regard.

src/commands/cronjob/update.tsx
niristius: leftover from a different active/inactive implementation.
gone now

src/lib/resources/cronjob/timeout.ts
```diff
@@ -0,0 +1,7 @@
+export function checkTimeout(timeout: number) {
+  if (timeout > 36800) {
```
martin-helmich, line 2: 36800 seconds is 10.2 hours, not 24. This should be 86400.

Or going even further, I'd recommend sticking with the specialized data type as long as possible:

```suggestion
const maxDuration = Duration.fromString("24h");

export function checkTimeout(timeout: Duration) {
  if (duration.compare(maxDuration) > 0) {
```

src/commands/cronjob/update.tsx
```diff
@@ -0,0 +1,127 @@
...
+      email,
+      timeout,
+      url,
+      interpreter,
+      command,
+      enable,
+      disable,
+    } = this.flags;
+
+    timeout && checkTimeout(timeout.seconds);
```
martin-helmich, line 89:
```suggestion
    if (timeout) {
        checkTimeout(timeout.seconds);
    }
```

src/commands/cronjob/create.tsx
```diff
@@ -97,18 +87,28 @@ export class Create extends ExecRenderBaseCommand<typeof Create, Result> {
       throw new Error("no project found for app installation");
     }
 
+    const destination: CronjobCreationData["destination"] =
+      buildCronjobDestination(url, command, interpreter) ??
+      (() => {
+        throw new Error(
+          "Url or command and interpreter combination could not be parsed.",
+        );
+      })();
```
martin-helmich, line 96: This looks a bit awkward. Maybe move the `throw` statement directly to `buildCronjobDestination`?

martin-helmich (changes requested):

niristius (commented):
short inquiry

src/commands/cronjob/create.tsx
```diff
@@ -65,6 +73,8 @@ export class Create extends ExecRenderBaseCommand<typeof Create, Result> {
       timeout,
     } = this.flags;
 
+    checkTimeout(timeout.seconds);
```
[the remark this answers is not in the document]
martin-helmich: I'm conflicted. On the one hand, this is a useful usability optimization. On the other hand, this is easy to run out-of-sync, if the backend validation ever changes without taking the frontend validation into account.

I think I would lean more towards leaving this kind of validation entirely in the backend, if only to have it all in one place.

src/commands/cronjob/update.tsx
```diff
@@ -77,54 +86,39 @@ export default class Update extends ExecRenderBaseCommand<
...
+      ...(destination !== undefined && {
+        destination,
+      }),
+      ...(enable && { active: true }),
+      ...(disable && { active: false }),
+      ...(description && { description }),
+      ...(timeout && { timeout: timeout.seconds }),
+      ...(email && { email }),
+      ...(interval && { interval }),
+    };
```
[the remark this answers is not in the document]
martin-helmich: Personally, I find this harder to read (even if shorter) than the chain of `if`s before.

On a side note: If it's a simple matter of "if undefined, omit it from the payload", you can still simply (unconditionally) set `description: description`; if the value is set to `undefined`, it will be omitted from the generated JSON in any case (set it to `null`, if you explicitly want it to be in the JSON body).

This should be functionally equivalent:

```typescript
const updateCronjobPayload: CronjobUpdateData = {
  destination,
  active: enable ? true : (disable ? false : undefined), // note: maybe keep the if/else construct instead of the nested ternary for readability
  description,
  timeout: timeout?.seconds, // will evaluate to "undefined" when "timeout" is undefined
  email,
  interval,
};
```

src/commands/cronjob/create.tsx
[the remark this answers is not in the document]
niristius: Ill make efforts to get the api error to be more helpful.

src/commands/cronjob/create.tsx
```diff
@@ -87,23 +79,18 @@ export class Create extends ExecRenderBaseCommand<typeof Create, Result> {
...
-        data: cronjobCreationData,
+        data: {
+          appId: appInstallationId,
+          active: !disable,
+          description,
+          interval,
+          email,
+          destination: buildCronjobDestination(url, command, interpreter),
+          timeout: timeout.seconds,
+        },
```
niristius, line 93: For some reason I was under the impression that if a key has an undefined value, currently set values for conjobs (or whichever entity is being updated) will be overwritten, I played around for a bit and it seems to not be the case wich makes this approach the most concise.

src/commands/cronjob/update.tsx
```diff
@@ -86,28 +81,7 @@ export default class Update extends ExecRenderBaseCommand<
       disable,
     } = this.flags;
 
-    if (timeout) {
-      checkTimeout(timeout);
-    }
```
niristius: I removed the check. Im making an effort to get a more helpful error message from the api or be able to read limits from the spec and use them in the client to be able to use checks that sync themselves.

niristius (commented):
final touches

martin-helmich (approved):

PR #488, opened 2024-05-14
Display more details for API errors

martin-helmich:
- **Tread Axios errors as API client errors**
- **Display more error details on API errors**


martin-helmich:
This would have saved me a lot of unnecessary debugging time with stuff like https://github.com/mittwald/ddev/actions/runs/9050065010 😅

PR #520, opened 2024-05-29
Add "app update" command to modify entrypoint/documentroot/...

martin-helmich:
- **Add "app update" command to change entrypoint+docroot**
- **Update README**

### DIFF
--- a/src/commands/app/upgrade.tsx
+++ b/src/commands/app/upgrade.tsx
@@ -27,7 +27,6 @@ import { MittwaldAPIV2, MittwaldAPIV2Client } from "@mittwald/api-client";
 import { waitUntilAppStateHasNormalized } from "../../lib/resources/app/wait.js";
 import { assertStatus } from "@mittwald/api-client-commons";
 import { waitFlags } from "../../lib/wait.js";
-import assertSuccess from "../../lib/apiutil/assert_success.js";
 import { ProcessFlags } from "../../rendering/process/process_flags.js";
 import semver from "semver/preload.js";
 


src/commands/app/upgrade.tsx
```diff
@@ -221,3 +260,80 @@ async function forceTargetVersionSelection(
...
+      systemSoftwareVersion: string | undefined;
+    };
+  } = {};
+  systemSoftware[dependencySoftware.data.id as string] = {
+    systemSoftwareVersion: dependencyTargetVersion.id,
+  };
+
+  const data = { systemSoftware };
+
+  const dependencyUpgradeResponse = await apiClient.app.patchAppinstallation({
```
martin-helmich, line 333: Doesn't this result in a multitude of `patchAppinstallation` requests (once for the upgrade itself, and then once for each dependency)? Might it be more robust to aggregate all those in to a single `patchAppinstallation` request?

niristius: Yeah it does, I changed the implementation to aggregate everything into one request.

niristius:
some testing was done although due rather current versions of everything in the studio, the amount of dependencies that need to be upgraded are very limited. I could only force PHP version to be low enough to require an upgrade during the app upgrade process.

niristius:
For clarificatrion - I just realized this.
I kinda hijacked this branch.
@martin-helmich initially add the `app update` command to change configurations of an app, entrypoints to be precise.
What I added is the `app upgrade` command to trigger version upgrades for apps.
This is ready to be merged.

src/commands/app/upgrade.tsx
```diff
@@ -227,3 +274,61 @@ async function forceTargetVersionSelection(
...
+        dependencyVersion.internalVersion,
+        dependencyTargetVersion.internalVersion,
+      )
+    ) {
+      dependencyTargetVersion = dependencyVersion;
+    }
+  }
+
+  if (!force) {
+    const confirmed: boolean = await process.addConfirmation(
```
martin-helmich, line 317: Possible UX optimization: Maybe only prompt for confirmation _once_, for all proposed system software updates, and not once for each?

src/commands/app/upgrade.tsx
```diff
@@ -227,3 +274,61 @@ async function forceTargetVersionSelection(
...
+) {
+  const dependencySoftware = await apiClient.app.getSystemsoftware({
+    systemSoftwareId: dependency.systemSoftwareId,
+  });
+  assertSuccess(dependencySoftware);
+
+  const dependencyVersionList = await apiClient.app.listSystemsoftwareversions({
+    systemSoftwareId: dependency.systemSoftwareId,
+  });
+  assertSuccess(dependencyVersionList);
```
martin-helmich, line 293: This API endpoint natively supports a `versionRange` query parameter,[^1] which I'd recommend using. By relying on the backend for this, you're guaranteed to use the same version selection process that the mStudio uses.

Then there's also the `recommended` filter (about which you probably know more than me), which you could use to further filter the selection of system software versions (I don't know if it's possible for the endpoint to return ≠1 recommended versions, though).

[^1]: https://developer.mittwald.de/docs/v2/reference/app/app-list-systemsoftwareversions/

src/commands/app/upgrade.tsx
```diff
@@ -227,3 +274,61 @@ async function forceTargetVersionSelection(
...
+
+  let dependencyTargetVersion: AppSystemSoftwareVersion = {
+    id: "not yet set",
```
martin-helmich, line 296: Maybe add an assertion after the `for` loop that a target version was actually set (considering the case that the loop was completed with no version matching the selector)?

src/commands/app/upgrade.tsx
niristius: Oh very good point, should've looked deeper into this.

The recommended filter will limit the returned software version to a certain amount of major, minor or patchlevel versions. For the example of PHP version it will return the latest three minor versions if they are within the version range.

Thus I'll leave the rest of this function as is, minus the semver check since that's redundant.

niristius:
@martin-helmich I cant request a review from you since I hijacked this branch from you. So here's my written request 😄

PR #571, opened 2024-06-14
add humhub as installable and upgradeable Software

niristius:


niristius:
Somehow `yarn readme:generate` threw some newlines into the mix which were not there before?
for example:
``` 
 -  -p, --project-id=<value>  ID or short ID of a project; this flag is optional if a default project is set in the context
+  -p, --project-id=<value>
+
+    ID or short ID of a project; this flag is optional if a default project is set in the context
 
     May contain a short ID or a full ID of a project; you can also use the "mw context set --project-id=<VALUE>" command
     to persistently set a default project for all commands that accept this flag.
```
if I read the linter correctly. but the yarn scripts have not been touched?

martin-helmich:
> Somehow `yarn readme:generate` threw away some newlines? for example:
> 
> ```
>  -  -p, --project-id=<value>  ID or short ID of a project; this flag is optional if a default project is set in the context
> +  -p, --project-id=<value>
> +
> +    ID or short ID of a project; this flag is optional if a default project is set in the context
>  
>      May contain a short ID or a full ID of a project; you can also use the "mw context set --project-id=<VALUE>" command
>      to persistently set a default project for all commands that accept this flag.
> ```
> 
> if I read the linter correctly. but the yarn scripts have not been touched?

I think this might be related to the `oclif` dependency having been bumped recently; have you made sure your node_modules are up-to-date?

PR #653, opened 2024-07-15
feat: add mail deliverybox

tillwestpfahl:
Adds all missing delivery box routes to the CLI.


martin-helmich:
@tillwestpfahl Is this ready for review?

tillwestpfahl:
Nope, I was waiting for the Readme-Issue to be fixed to continue with this.

src/commands/mail/deliverybox/create.tsx
```diff
@@ -0,0 +1,153 @@
...
+    This command can be used to create a new mail deliverybox in a project.
+  
+    When running this command with the --quiet flag, the output will contain the ID of the newly created deliverybox.
+    In addition, when run with --generated-password the output will be the ID of the newly created deliverybox, followed by a tab character and the generated password.`;
+  static flags = {
+    ...projectFlags,
+    ...processFlags,
+    description: Flags.string({
+      char: "d",
+      summary: "mail devliverybox description",
```
martin-helmich, line 37:
```suggestion
      summary: "mail deliverybox description",
```

src/commands/mail/deliverybox/create.tsx
```diff
@@ -0,0 +1,153 @@
...
+      description:
+        "This flag will cause the command to generate a random 32-character password for the deliverybox; when running with --quiet, the devliverybox ID and the password will be printed to stdout, separated by a tab character.",
+    }),
+  };
+
+  static examples = [
+    {
+      description: "Create non-interactively with password",
+      command:
+        "$ read -s PASSWORD &&\n <%= config.bin %> <%= command.id %> --password $PASSWORD --description my personal deliverybox",
```
martin-helmich, line 56:
```suggestion
        "$ read -s PASSWORD &&\n <%= config.bin %> <%= command.id %> --password $PASSWORD --description 'my personal deliverybox'",
```

src/commands/mail/deliverybox/create.tsx
```diff
@@ -0,0 +1,153 @@
...
+    },
+    {
+      description: "Create non-interactively with random password",
+      command:
+        "<%= config.bin %> <%= command.id %> --random-password --description my personal deliverybox",
```
martin-helmich, line 61:
```suggestion
        "<%= config.bin %> <%= command.id %> --random-password --description 'my personal deliverybox'",
```

src/commands/mail/deliverybox/create.tsx
```diff
@@ -0,0 +1,153 @@
...
+    ...projectFlags,
+    ...processFlags,
+    description: Flags.string({
+      char: "d",
+      summary: "mail devliverybox description",
+      required: true,
+    }),
+    password: Flags.string({
+      summary: "deliverybox password",
+      description:
```
martin-helmich, line 42: Use the `exclusive` property (IIRC, refer to documentation) to configure that _either_ `--password` _or_ `--random-password` (or none of them) is supplied.

src/commands/mail/deliverybox/create.tsx
```diff
@@ -0,0 +1,153 @@
...
+        },
+      );
+
+      process.addInfo(
+        <Text>
+          generated password: <Value>{generated}</Value>
+        </Text>,
+      );
+      return [generated, true];
+    }
```
martin-helmich, line 86: This (together with a few other bits and pieces in this class) is probably a duplication of the `mail address create` command? Can we extract the duplicated code to a commonly-used function?

martin-helmich (changes requested):

src/commands/mail/commons.tsx
martin-helmich, line 1: Can we move this file somewhere to `lib`, maybe to `lib/resources/mail`? Typically, each file in `commands` contains an actual command class, with library code being placed in `lib`, so this might be confusing.

src/commands/mail/address/create.tsx
```diff
@@ -58,11 +57,13 @@ export default class Create extends ExecRenderBaseCommand<
     }),
     password: Flags.string({
       summary: "mailbox password",
+      exclusive: ["random-password"],
```
martin-helmich, line 60: Good catch! :+1:

src/lib/resources/mail/flags.ts
```diff
@@ -26,3 +24,27 @@ export const {
...
+
+    const response = await apiClient.mail.listDeliveryBoxes({ projectId });
+    assertStatus(response, 200);
+
+    const deliveryBox = response.data.find(
+      (deliveryBox) => deliveryBox.id === id,
+    );
+    if (!deliveryBox) {
+      throw new Error(`mail delivery box with id "${id}" not found`);
+    }
```
martin-helmich, line 46: This should be unnecessary IMO; effectively, all this does is looking up a delivery box based on its ID, which you can also do using a `getDeliveryBox` operation (which works even better, because that call also can resolve short IDs (at least potentially, because AFAIK that's not implemented yet for mail addresses and delivery boxes).

TLDR; You can probably omit the `normalize` function entirely (not sure if there's a default implementation), because all this function would do is just return the same ID that's passed into it.

martin-helmich (changes requested):

src/lib/resources/mail/flags.ts
```diff
@@ -26,3 +24,27 @@ export const {
...
+
+    return deliveryBox.id;
+  },
+});
```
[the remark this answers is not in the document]
martin-helmich: Yes, using `makeProjectFlagSet` would make sense for a short-id lookup -- and that only in cases where it's not supported natively by the API. Currently it's used for domain/DNS names (when a project is set in the context, you can use a domain name instead of the domain ID).

martin-helmich (approved):

PR #698, opened 2024-08-13
Implement mysql user creation and update

niristius:
~This is currently blocked by the lack of withMySQLId and withMySQLUserId functions, both are in the works~
this is outdated information, they are not required


src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+      options: ["readonly", "full"],
+    }),
+    description: Flags.string({
+      required: true,
+      description: "Description of the mysql user",
+    }),
+    password: Flags.string({
+      required: true,
+      description: "Password used for authentication",
+      exactlyOne: ["public-key", "password"],
```
martin-helmich, line 41: There is no `public-key` flag. Copy&paste error?

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+    "database-id": Args.string({
+      description: "ID of the MySQL Database to create a user for",
+    }),
+  };
+  static flags = {
+    ...processFlags,
+    "database-id": Flags.string({
+      required: true,
+      description: "ID of the MySQL Database to create a user for",
+    }),
```
martin-helmich, line 28: For a single command, there is no need to have BOTH an argument and a flag for the same value. Pick one (to stay consistent with other commands like `database mysql create`, which accepts a `--project-id` _flag_, this should probably be a flag).

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+    "access-level": Flags.string({
+      required: true,
+      description: "Access level for this mysql user",
```
martin-helmich, line 31: Inconsistent spelling of `MySQL` across all descriptions; I'd favor sticking with `MySQL` instead of `mysql` because that's the official product name.

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+      databaseId: mysqlDatabaseId,
+      description,
+      password,
+    };
+
+    if (password.length < 12) {
+      throw new Error(
+        "Your chosen password is too short. Please choose a secure password with at least 12 characters.",
+      );
+    }
```
martin-helmich, line 96: Password policies should be enforced in the backend (and without a mechanism to synchronize password policies across clients, preferably _only_ in the backend).

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+    let mysqlDatabaseId: string = "";
+    if (this.flags["database-id"]) {
+      mysqlDatabaseId = this.flags["database-id"];
+    } else if (this.args["database-id"]) {
+      mysqlDatabaseId = this.args["database-id"];
+    } else {
+      await process.error(
+        "Please provide a backup schedule id as flag or argument",
+      );
+    }
```
martin-helmich, line 68: What should `withMySQLId` do in this context? Setting a MySQL database id in the CLI context is currently not supported anyway (and I wouldn't really deem it necessary), and the lookup from short IDs to UUIDs is supported in the API now, anyway.

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,146 @@
...
+      "external-access": externalAccess,
+    } = this.flags;
+
+    const createMysqlUserPayload: {
+      accessLevel: "full" | "readonly";
+      databaseId: string;
+      description: string;
+      password: string;
+      accessIpMask?: string | undefined;
+      externalAccess?: boolean | undefined;
```
martin-helmich, line 84: Can we extract these types somewhere from the API client? Or are they inlined there, which makes this duplication necessary?

martin-helmich (changes requested):

src/commands/database/mysql/user/create.tsx
niristius: I was under the impression, the with[entity]Id functions where there to handle the handover of the ID via args or flags, just to make usage more convinient. That's also what I mimicked with my implementation. I will revise this and stick to the flag.

src/commands/database/mysql/user/create.tsx
niristius: sloppy, fixed.

src/commands/database/mysql/user/create.tsx
niristius: sticking with the flag

src/commands/database/mysql/user/create.tsx
niristius: ✅

src/commands/database/mysql/user/create.tsx
niristius: Will stick with MySQL for all mentions

src/commands/database/mysql/user/create.tsx
niristius: This I would be very interested in learning more about. 
The declaration of the type there helps to guide the creation of the Payload wich is then sent to the api.
these type I have copied from the data key of the (in this case) createMysqlUser function.

Two things though:
1.  since I only need the data key, the type would look something like 
```type MyQSLUserData = MittwaldAPIV2.Components.Schemas.DatabaseMySqlUser.data;```
which is syntactically not possible

```TS2713: Cannot access DatabaseMySqlUser.data because DatabaseMySqlUser is a type, but not a namespace. Did you mean to retrieve the type of the property data in DatabaseMySqlUser with 'DatabaseMySqlUser[data]'?```

If I try to retrieve the data property though, using

``

Comment (martin-helmich) [reply src/commands/database/mysql/user/create.tsx:84]: Bleh, there actually is no named type for this method's request body.
This is hacky AF, but you can (just tried it out with `createMySQLDatabase`) do this:

```typescript
type CreateParam = Parameters<MittwaldAPIV2Client["database"]["createMysqlDatabase"]>[0]["data"];
```

martin-helmich: See https://github.com/mittwald/cli/pull/720, for example.

martin-helmich: Addendum: Using helpers like `Partial<?>`, `Omit<?>` or `Required<?>`,[^1] you can also modify the inferred types, to make properties optional or required, or omit them entirely.

[^1]: https://www.typescriptlang.org/docs/handbook/utility-types.html

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,126 @@
...
+        : (currentMysqlUserData.data.accessIpMask as string),
+      externalAccess: externalAccess
+        ? externalAccess
+        : (currentMysqlUserData.data.externalAccess as boolean),
+    };
+
+    if (Object.keys(updateMysqlUserPayload).length == 1) {
+      await process.complete(
+        <Success>Nothing to change. Have a good day!</Success>,
+      );
```
martin-helmich, line 92: Seems like there should be a `return` here?

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,126 @@
...
+          await this.apiClient.database.updateMysqlUserPassword({
+            mysqlUserId,
+            data: {
+              password,
+            },
+          });
+        assertSuccess(updatePasswordResponse);
+      });
+
+      if (accessLevel || description || accessIpMask || externalAccess) {
```
martin-helmich, line 107: This is still within the `if (password) {` clause. Shouldn't it be outside of the previous condition?

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,126 @@
...
+  typeof Update,
+  UpdateResult
+> {
+  static description = "Create a new mysql user";
+  static flags = {
+    ...processFlags,
+    "mysql-user-id": Flags.string({
+      required: true,
+      description: "MySQL User ID of the user to be updated",
+    }),
```
martin-helmich, line 26: Other commands that directly affect a specific resource use that resource ID as argument (comparison: all `delete` and `get` commands, and also other update commands like `mail address update`)

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,126 @@
...
+    "access-level": Flags.string({
+      description: "Access level for this MySQL user",
+      options: ["readonly", "full"],
+    }),
```
martin-helmich, line 30: You can strictly type these flags by using `Flags.custom<"full" | "readonly">({...})()`. See `database mysql compare` for comparison. You can also use `Flags.custom` to define a common flag definition (somewhere in `lib/resources/database/mysql`) and reuse that flag definition in all required places.

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,126 @@
...
+    description: Flags.string({
+      description: "Description of the MySQL user",
+    }),
+    password: Flags.string({
+      description: "Password used for authentication",
+    }),
+    "access-ip-mask": Flags.string({
+      description: "IP from wich external access will be exclusively allowed",
```
martin-helmich, line 38:
```suggestion
      description: "IP from which external access will be exclusively allowed",
```

martin-helmich (changes requested):

src/commands/database/mysql/user/update.tsx
niristius: very much so, yes.

src/commands/database/mysql/user/update.tsx
```diff
@@ -0,0 +1,131 @@
...
+type UpdateResult = void;
+type MyQSLUserUpdateData = Parameters<
+  MittwaldAPIV2Client["database"]["updateMysqlUser"]
+>[0]["data"];
+
+export default class Update extends ExecRenderBaseCommand<
+  typeof Update,
+  UpdateResult
+> {
+  static description = "Create a new mysql user";
```
martin-helmich, line 20:
```suggestion
  static description = "Updates an existing MySQL user";
```

niristius:
something important to state:

I edited the `database mysql user list` command to display user uuids.
Currently it is not possible to use the database username as ID.
This is being reviewed internally but the outcome is unclear sind it technically is not a shortId.
my wish is to just the username for everything, since it is unique aswell.
Until then I implimented this as a workaround

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,110 @@
...
+  static description = "Create a new MySQL user";
+  static flags = {
+    ...processFlags,
+    "database-id": mysqlUserFlagDefinitions["database-id"]({ required: true }),
+    "access-level": mysqlUserFlagDefinitions["access-level"]({
+      required: true,
+    }),
+    description: mysqlUserFlagDefinitions.description({ required: true }),
+    password: mysqlUserFlagDefinitions.password({ required: true }),
+    "access-ip-mask": mysqlUserFlagDefinitions["access-ip-mask"](),
```
martin-helmich, line 32: Should only be set when `--enable-external-access` is also set.

```suggestion
    "access-ip-mask": mysqlUserFlagDefinitions["access-ip-mask"]({
      dependsOn: ["enable-external-access"]
    }),
```

martin-helmich (approved):

src/commands/database/mysql/user/create.tsx
```diff
@@ -0,0 +1,112 @@
...
+  static flags = {
+    ...processFlags,
+    "database-id": mysqlUserFlagDefinitions["database-id"]({ required: true }),
+    "access-level": mysqlUserFlagDefinitions["access-level"]({
+      required: true,
+    }),
+    description: mysqlUserFlagDefinitions.description({ required: true }),
+    password: mysqlUserFlagDefinitions.password({ required: true }),
+    "access-ip-mask": mysqlUserFlagDefinitions["access-ip-mask"]({
+      dependsOn: ["enable-external-access"]
```
martin-helmich, line 33: Thanks for nothing, eslint 😖 

```suggestion
      dependsOn: ["enable-external-access"],
```

Issue #654, opened 2024-07-15
yarn generate:readme with inconsistent results


tillwestpfahl:
> [niristius](https://github.com/niristius) commented [4 days ago](https://github.com/mittwald/cli/pull/650#issue-2402989189)
> So I deleted the old branch since it was kinda stale and I am still trying to understand Node, yarn and oclif a bit more.
> And since the changes are rather trivial I started over.
> 
> Sadly the yarn generate:readme command still adds newlines where they dont belong.
> 
> I tried
> 
> using Node 18, 20, 22
> using oclif 4.14, 4.13, 4.12, 4.10, 3.17, 3.16, 3.15, 3.14
> still no cigar.
> @martin-helmich please advise

I have the same exact problem ([see here](https://github.com/mittwald/cli/actions/runs/9938894314/job/27452328300)). Also for some reason, the generated readme looks entirely different after a fresh repo clone.

Also `yarn format` changes the README formatting yet another way (which to be fair, is probably expected).

niristius:
This PR is stuggling with this aswell: https://github.com/mittwald/cli/pull/465

niristius:
And this one aswell: https://github.com/mittwald/cli/pull/687

niristius:
and also these
https://github.com/mittwald/cli/pull/699
https://github.com/mittwald/cli/pull/700

martin-helmich:
Did some digging on this; apparently, the output differs, depending on whether you're invoking `yarn generate:readme` or calling `./node_modules/.bin/oclif readme` directly. Will investigate further. 🕵️‍♂️

PR #700, opened 2024-08-13
Add magento installation command

niristius:


niristius:
blocked by https://github.com/mittwald/cli/issues/654

src/lib/resources/app/flags.tsx
```diff
@@ -322,7 +338,11 @@ export async function autofillFlags(
 
   // Shop Language Code
   if (necessaryFlags.includes("shop-lang") && !flags["shop-lang"]) {
-    flags["shop-lang"] = "de-DE";
+    if (appName.toLowerCase().includes("magento")) {
```
martin-helmich, line 341: This seems like a code smell; adding exceptions for special cases in a generic module such as this has the potential to become an unmaintainable mess in the future. Maybe we can add the option to specify default/fallback values in the `AppInstaller` instance, without having to hardcode them in the `buildFlagsWithDescription` function (maybe `buildFlagsWithDescription` could accept the default values as a parameter, which gets passed in from the `AppInstaller`?)

martin-helmich (changes requested):

martin-helmich (approved):

PR #701, opened 2024-08-14
Backup schedule operations

niristius:


src/commands/backup/schedule/create.tsx
```diff
@@ -0,0 +1,100 @@
...
+      description: "Set the description for the backup schedule",
+    }),
+    schedule: Flags.string({
+      required: true,
+      description: "Define the schedule itself",
+    }),
+    ttl: Flags.string({
+      required: true,
+      description:
+        "Define the backup storage period in days, for through this schedule created backups",
```
martin-helmich, line 33: From what I read in the (albeit not very thoroughly documented) API spec[^1], this is _not_ just a "number of days", but a generic interval string, in which the user can specify the unit themselves (so for example, `12d`, `3h` or `3w`). This should be documented accordingly.

Also, check if you can use the `Duration` class,[^2] which also offers a specialized flag type (its `toString` method might need to be extended though, to fit this use case better).

[^1]: https://developer.mittwald.de/docs/v2/reference/backup/backup-create-project-backup-schedule/
[^2]: https://github.com/mittwald/cli/blob/master/src/lib/units/Duration.ts

src/commands/backup/schedule/create.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }),
+  };
+
+  protected async exec(): Promise<Result> {
+    const process = makeProcessRenderer(this.flags, "Creating a new sftp User");
```
martin-helmich, line 38:
```suggestion
    const process = makeProcessRenderer(this.flags, "Creating a new backup schedule");
```

src/commands/backup/schedule/create.tsx
```diff
@@ -0,0 +1,100 @@
...
+      schedule: string;
+      ttl: string;
+    } = {
+      schedule,
+      ttl,
+    };
+
+    if (description) {
+      createBackupSchedulePayload.description = description;
+    }
```
martin-helmich, line 53: If `description` is of type `string|undefined` anyway, this could be simplified to:

```suggestion
    const createBackupSchedulePayload: {
      description?: string | undefined;
      schedule: string;
      ttl: string;
    } = {
      description,
      schedule,
      ttl,
    };
```

martin-helmich (changes requested):

src/commands/backup/schedule/create.tsx
niristius: I was under the impression that if I PATCH a parameter as "undefined" for an existing Parameter, it will get purged.
Meaning if it doesn't contain a value but the property is set in the Studio, after the Request its gone.
That's why I seperated optional parameters - so that they aren't a part of the update payload unless they are set.

If my assumprion is wrong I'll gladly simplify the code. I just want to make sure.

src/commands/backup/schedule/create.tsx
niristius: I was playing around with the values but anything except "[int]d" yields the response 
```
 "type": "ValidationError",
    "message": "3 INVALID_ARGUMENT: ttl format 7 is not allowed, allowed is '{days}d' e.g. '7d', '30d'",
    "validationErrors": [
```
Now Im sure that we could take on (milli-)seconds as ttl format in true unix fashion, I want to put to discussion though if the pros outweight the cons toreason for implementing it. In my opinion `-ttl=[int]` whould be all a user has to do for the ttl.

there are even hard limits in place:
```
"type": "ValidationError",
"message": "3 INVALID_ARGUMENT: ttl 500d with value 500 is not allowed, allowed values are between 7 and 400",
```

niristius:
I did extensive tests and this PR is ready as soon as the ttl conversation is resolved

src/commands/backup/schedule/delete.tsx
```diff
@@ -0,0 +1,44 @@
...
+      description: "ID of schedule to delete",
+      required: true,
+    }),
+  };
+  static flags = {
+    ...processFlags,
+  };
+
+  protected async deleteResource(): Promise<void> {
+    const process = makeProcessRenderer(this.flags, "Updating backup schedule");
```
martin-helmich, line 23:
```suggestion
    const process = makeProcessRenderer(this.flags, "Deleting backup schedule");
```

src/commands/backup/schedule/update.tsx
```diff
@@ -0,0 +1,80 @@
...
+import type { MittwaldAPIV2Client } from "@mittwald/api-client";
+import { backupScheduleFlagDefinitions } from "../../../lib/resources/backup/schedule/flags.js";
+
+type UpdateResult = void;
+type backupScheduleUpdatePayload = Parameters<
+  MittwaldAPIV2Client["backup"]["updateProjectBackupSchedule"]
+>[0]["data"];
+
+export default class Create extends ExecRenderBaseCommand<
+  typeof Create,
```
martin-helmich, line 19:
```suggestion
export default class Update extends ExecRenderBaseCommand<
  typeof Update,
```

src/lib/resources/backup/schedule/flags.ts
```diff
@@ -0,0 +1,21 @@
...
+    description:
+      "Must be specified as a cron schedule expression. " +
+      "Cannot be scheduled more often than once per hour. " +
+      "Defines the interval at which the backup creation will be executed.",
+  }),
+  ttl: Flags.custom<string>({
+    summary: "Define the backup retention period in days for backups created.",
+    description:
+      "Must be specified as a natural number between 7 and 400, representing the number of days the backup will be kept.",
+  }),
```
martin-helmich, line 20: Refresh my memory: Is the format like `7d` required? If so, this should be mentioned in the description. If not, this should better be of type `number`.

niristius: format `[amount]d` is correct. will adapt the description

niristius: come to think of it though the cli could take the guesswork off the user’s hands for whom the input as a pure number might be more convinient. But that will add some magic to the cli wich might be contradictionary to when talking directly to the api.

src/commands/backup/schedule/delete.tsx
```diff
@@ -0,0 +1,44 @@
...
+        "The system backup created through this schedule is reserved to restore your project in an emergency. " +
+          "It can not be deleted.",
+      );
+    }
+
+    const response = await this.apiClient.backup.deleteProjectBackupSchedule({
+      projectBackupScheduleId,
+    });
+
+    assertStatus(response, 204);
```
martin-helmich, line 42: Better use `assertSuccess` in this case; the handling of 204 status codes has been notoriously inconsistent in the past, and there have been (technically breaking) changes where a 204 status code has been changed to a 200 status code.

martin-helmich (changes requested):

martin-helmich (approved):

PR #744, opened 2024-09-03
Add option to only up/download a subdirectory of an app installation

martin-helmich:
This PR adds a flag to the `app download` and `app upload` command to restrict rsync to a specific sub directory on the remote side.

This is useful when you only want to up/download a specific subdirectory of the app installation, for example when you are using a deployment tool that manages the app installation directory itself, and you only want to up/download exempt files, like environment specific configuration files or user data.

### DIFF
--- a/docs/app.md
+++ b/docs/app.md
@@ -382,6 +382,15 @@ DESCRIPTION
   rsync. Have a look at https://manpages.ubuntu.com/manpages/noble/en/man1/rsync.1.html#filter%20rules for more
   information on how to write filter rules.
 
+EXAMPLES
+  Download entire app to current working directory
+
+    $ mw app download .
+
+  Download only shared dir from a deployer-managed app
+
+    $ mw app download --remote-sub-directory=shared .
+
 FLAG DESCRIPTIONS
   -q, --quiet  suppress process output and only display a machine-readable summary.
 
@@ -392,7 +401,8 @@ FLAG DESCRIPTIONS
 
     This is particularly useful when you only want to download a specific sub-directory of the app installation, for
     example when you are using a deployment tool that manages the app installation directory itself, and you only want
-    to download exempt files, like environment specific configuration files or user data.
+    to download exempt files, like environment specific configuration files or user data. For example, if you want to
+    download from "/html/my-app-XXXXX/config", set "--remote-sub-directory=config".
 
   --ssh-identity-file=<value>  the SSH identity file (private key) to use for public key authentication.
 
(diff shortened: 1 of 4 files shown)


tehplague (approved):

tehplague (approved):

docs/app.md
```diff
@@ -352,18 +352,19 @@ Download the filesystem of an app within a project to your local machine
...
-      --delete              delete local files that are not present on the server
-      --dry-run             do not actually download the app installation
-      --exclude=<value>...  [default: ] exclude files matching the given pattern
-      --target=<value>      (required) target directory to download the app installation to
+  -q, --quiet                  suppress process output and only display a machine-readable summary.
+      --delete                 delete local files that are not present on the server
+      --dry-run                do not actually download the app installation
+      --exclude=<value>...     [default: ] exclude files matching the given pattern
+      --sub-directory=<value>  specify a sub-directory within the app installation to download
+      --target=<value>         (required) target directory to download the app installation to
```
LukasFritzeDev, line 367: For a Moment I was confused why I can not set the `sub-directory` directly as part of `target` … then I noticed the difference:
* `target` is the local path on my machine
* `sub-directory` is a relative path on the remote machine

Maybe the flag should be called `remote-sub-dir` to better reflect this?

docs/app.md
```diff
@@ -407,6 +408,12 @@ FLAG DESCRIPTIONS
     will be used for this.
 
     You can also set this value by setting the MITTWALD_SSH_USER environment variable.
+
+  --sub-directory=<value>  specify a sub-directory within the app installation to download
+
+    This is particularly useful when you only want to download a specific sub-directory of the app installation, for
+    example when you are using a deployment tool that manages the app installation directory itself, and you only want
+    to download exempt files, like environment specific configuration files or user data.
```
LukasFritzeDev, line 416: It might be helpful to give an example. If I know the file hierarchy on the server, it might be helpful to have this in the description:

```suggestion
  --sub-directory=<value>  specify a sub-directory within the app installation to download

    This is particularly useful when you only want to download a specific sub-directory of the app installation, for
    example when you are using a deployment tool that manages the app installation directory itself, and you only want
    to download exempt files, like environment specific configuration files or user data. For example if you want to download `html/my-app-12345/config` set `--sub-directory=config`
```

docs/app.md
martin-helmich: Good catch. :+1:  I like `remote-sub-dir` for its better clarity; will rename...

docs/app.md
martin-helmich: Fixed.

docs/app.md
```diff
@@ -2064,6 +2074,13 @@ FLAG DESCRIPTIONS
     This flag controls if you want to see the process output or only a summary. When using mw non-interactively (e.g. in
     scripts), you can use this flag to easily get the IDs of created resources for further processing.
 
+  --remote-sub-directory=<value>  specify a sub-directory within the app installation to upload
+
+    This is particularly useful when you only want to upload a specific sub-directory of the app installation, for
+    example when you are using a deployment tool that manages the app installation directory itself, and you only want
+    to upload exempt files, like environment specific configuration files or user data. For example, if you want to
+    download "/html/my-app-XXXXX/config", set "--remote-sub-directory=config".
```
LukasFritzeDev, line 2082: > you only want to **upload** exempt files, […] if you want to download …

🤔

LukasFritzeDev (approved):

PR #750, opened 2024-09-04
Target/source of app upload|download command should be arg, not flag

### LINKED ISSUE #394: Streamline flags and args of up-/download commands
Why is this a flag and not an argument? I assumed that flags _should_ be optional an required input is given as argument.

_Originally posted by @LukasFritzeDev in https://github.com/mittwald/cli/pull/243#discussion_r1555480371_

martin-helmich:
This PR streamlines the flags+args of the `mw app upload|download` command.

Since these are required, it makes more sense to model them as positional arguments instead of flags.

Note that this change is very breaking, which is why we're introducing it before an eventual 1.0 release.

Fixes #394

### DIFF
--- a/docs/app.md
+++ b/docs/app.md
@@ -11,7 +11,7 @@ Manage apps, and app installations in your projects
 * [`mw app dependency list`](#mw-app-dependency-list)
 * [`mw app dependency update [INSTALLATION-ID]`](#mw-app-dependency-update-installation-id)
 * [`mw app dependency versions SYSTEMSOFTWARE`](#mw-app-dependency-versions-systemsoftware)
-* [`mw app download [INSTALLATION-ID]`](#mw-app-download-installation-id)
+* [`mw app download [INSTALLATION-ID] TARGET`](#mw-app-download-installation-id-target)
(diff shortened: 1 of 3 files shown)


docs/app.md
```diff
@@ -11,7 +11,7 @@ Manage apps, and app installations in your projects
 * [`mw app dependency list`](#mw-app-dependency-list)
 * [`mw app dependency update [INSTALLATION-ID]`](#mw-app-dependency-update-installation-id)
 * [`mw app dependency versions SYSTEMSOFTWARE`](#mw-app-dependency-versions-systemsoftware)
-* [`mw app download [INSTALLATION-ID]`](#mw-app-download-installation-id)
+* [`mw app download [INSTALLATION-ID] TARGET`](#mw-app-download-installation-id-target)
```
martin-helmich, line 14: Ah; now I remember why we didn't do this in the first place. This is not valid, because `[INSTALLATION-ID]` is optional any may not be followed by a required argument. So the only valid order would be `TARGET [INSTALLATION-ID]`, which also seems unintuitive. I'll just close this and leave this as-is.

Issue #922, opened 2024-11-25
Trigger App Installation with specific AppVersion-uuid


niristius:
**Is your feature request related to a problem? Please describe.**
To make Installation of Versions possible that are not returned by `apiClient.app.listAppversions`

**Describe the solution you'd like**
be able to pass the appVersion uuid with the `app install [software] --version=[uuid]`

**Describe alternatives you've considered**
Alternatively hidden version have to be exposed via the api directly which might not always be desireable (for example typo3 10 non-elts versions)

martin-helmich:
https://github.com/mittwald/cli/pull/934 has already been merged; closing this for now.

PR #934, opened 2024-11-28
let app and upgrade command accept an AppVersion uuid as target version

niristius:
this is my implementation to realize the requested feature from https://github.com/mittwald/cli/issues/922

### DIFF
--- a/src/commands/app/upgrade.tsx
+++ b/src/commands/app/upgrade.tsx
@@ -29,6 +29,7 @@ import { assertStatus } from "@mittwald/api-client-commons";
 import { waitFlags } from "../../lib/wait.js";
 import { ProcessFlags } from "../../rendering/process/process_flags.js";
 import semver from "semver/preload.js";
+import { validate as validateUuid } from "uuid";
 
 type AppApp = MittwaldAPIV2.Components.Schemas.AppApp;
 type AppAppInstallation = MittwaldAPIV2.Components.Schemas.AppAppInstallation;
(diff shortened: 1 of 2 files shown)


niristius:
upgrades are included too

src/lib/resources/app/Installer.tsx
```diff
@@ -91,11 +94,29 @@ export class AppInstaller<TFlagName extends AvailableFlagName> {
...
+    var appVersion: AppAppVersion;
+
+    if (validateUuid(appVersionId)) {
+      appVersion = await getAppVersionFromUuid(
+        apiClient,
+        this.appId,
+        appVersionId,
+      );
+    } else {
+      appVersion = await normalizeToAppVersionUuid(
```
martin-helmich, line 108: Maybe build the "if it looks like a UUID, just use that"-logic directly into `normalizeToAppVersionUuid`? IIRC, all other `normalize<whatever>` functions that we have floating around work more or less similar, anyway.

src/commands/app/upgrade.tsx
```diff
@@ -114,7 +115,19 @@ export class UpgradeApp extends ExecRenderBaseCommand<typeof UpgradeApp, void> {
...
+    ) {
+      targetAppVersion = await getAppVersionFromUuid(
+        this.apiClient,
+        currentApp.id,
+        this.flags["target-version"],
+      );
+    } else if (
+      this.flags["target-version"] &&
+      !validateUuid(this.flags["target-version"])
+    ) {
```
martin-helmich, line 130: These `AND`ed conditions could be deconstructed to (maybe?) more readable nested conditions:

```
else if (this.flags["target-version"]) {
  if (validateUuid(this.flags["target-version"])) {
    // ...
  } else {
    // ...
  }
} // ...
```

Also, the way how `targetAppVersion` is getting more and more complex, with more and more branches in this condition. I would suggest extracting the entire logic of populating `targetAppVersion` into a function.

martin-helmich (changes requested):
Some refactoring suggetions; also, mind the linter issues.

PR #1140, opened 2025-04-22
feat: add commands for publishing extensions

martin-helmich:


src/commands/contributor/extension/publish.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {
+  extensionManifestArg,
+  parseExtensionManifest,
+} from "../../../lib/resources/extension/args_contributor.js";
+
+export default class Publish extends ExecRenderBaseCommand<
+  typeof Publish,
+  void
+> {
+  static description = "Publish an extension on the marketplace";
```
freisenhauer, line 17: The publishing of extensions will only work if verification has been requested and successfully completed beforehand.

I believe the corresponding route to request verification has not yet been published

src/commands/contributor/extension/withdraw.tsx
```diff
@@ -0,0 +1,58 @@
...
+  protected async exec(): Promise<void> {
+    const p = makeProcessRenderer(this.flags, "Withdrawing extension");
+
+    const { reason } = this.flags;
+    const manifest = await parseExtensionManifest(
+      this.args["extension-manifest"],
+    );
+    const { contributorId, id } = manifest;
+
+    await p.runStep("Withdrawing extension", async () => {
```
freisenhauer, line 41: Should we ask for confirmation somehow? That's a rather drastic step

src/commands/contributor/extension/init.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { extensionManifestArg } from "../../../lib/resources/extension/args_contributor.js";
+import { ExtensionManifest } from "../../../lib/resources/extension/manifest.js";
+import * as uuid from "uuid";
+import { writeFile } from "fs/promises";
+import yaml from "js-yaml";
+import { Value } from "../../../rendering/react/components/Value.js";
+import { pathExists } from "../../../lib/util/fs/pathExists.js";
+
+export default class Init extends ExecRenderBaseCommand<typeof Init, void> {
+  static summary = "Init a new extension manifest file";
```
freisenhauer, line 18: good idea tbh!

src/commands/contributor/extension/withdraw.tsx
martin-helmich: Good call. We should probably treat it the same way as a "delete" command, with interactive confirmation and a `--force` flag. 🤔

src/commands/contributor/extension/publish.tsx
martin-helmich: Can we rely on the API to respond with an appropriate and sensible error message in this case (in which case we could leave this as-is and just present the API error message), or should we catch that condition (successful verification) beforehand client-side?

freisenhauer: The API will likely respond with a reasonable error message. I just wanted to point out, the current flow will be interrupted by writing some mittwald guys to verify the extension (including starting the verification process) until the corresponding API route is published.

PR #1259, opened 2025-08-07
Allow AppVersionUuids to be used as UpgradeTargetVersions

mhopfner:
This change allows bypassing the `targetAppVersionCandidates` check, enabling upgrades to versions that are currently hidden.
To enhance upgrade flexibility and strengthen cooperation with our customers, this provides a mechanism for manually initiating upgrades to versions that are currently not listed.
However, the UUID of the target version must be known or communicated in advance.
It is still not possible to arbitrarily up- or downgrade installations.
Downgrades, non-existent versions, or versions belonging to a different app than the AppInstallation being upgraded is from, remain prohibited by the API.


src/commands/app/upgrade.tsx
```diff
@@ -114,6 +118,15 @@ export class UpgradeApp extends ExecRenderBaseCommand<typeof UpgradeApp, void> {
           currentApp.id,
           currentAppVersion.id,
         )) as AppAppVersion;
+    } else if (
+      this.flags["target-version"] &&
+      validateUuid(this.flags["target-version"])
+    ) {
```
martin-helmich, line 124: This is becoming rather long, and difficult to get your head around. Maybe extract that whole "determine target version" logic in to a function, from which you could just return early (which might also help unravel all these if/else constructs?)

martin-helmich:
I'm not seeing a diff any longer after merging `master` into this branch. Did your changes not make it into the new `determineTargetAppVersion` function?

martin-helmich (approved):

Issue #1322, opened 2025-08-28
Error while mw update command


NeoBlack:
I get an error while update the CLI tool:

```
#> mw update
@mittwald/cli: Updating CLI from 1.7.0 to 1.9.0-0317a3a... !
╭──────────────────────────────────────────────────────────────────────────────╮
│ ERROR                                                                        │
│                                                                              │
│ An error occurred while executing this command:                              │
│                                                                              │
│   Error: ENOENT: no such file or directory, open                             │
│   '/Users/xxxxxx/.local/share/mw/channel'                                    │
│                                                                              │
│ If you believe this to be a bug, please open an issue at                     │
│ https://github.com/mittwald/cli/issues/new.                                  │
╰──────────────────────────────────────────────────────────────────────────────╯

  ERROR STACK TRACE

  Please provide this when opening a bug report.

  Error: ENOENT: no such file or directory, open '/Users/xxxxxx/.local/share/mw/channel'
      at async open (node:internal/fs/promises:639:25)
      at async writeFile (node:internal/fs/promises:1216:14)
      at async Updater.update (file:///opt/homebrew/Cellar/mw/1.7.0/libexec/node_modules/@oclif/plugin-update/dist/update.js:213:9)
      at async Updater.runUpdate (file:///opt/homebrew/Cellar/mw/1.7.0/libexec/node_modules/@oclif/plugin-update/dist/update.js:107:17)
      at async UpdateCommand._run (/opt/homebrew/Cellar/mw/1.7.0/libexec/node_modules/@oclif/core/lib/command.js:181:22)
      at async Config.runCommand (/opt/homebrew/Cellar/mw/1.7.0/libexec/node_modules/@oclif/core/lib/config/config.js:456:25)
      at async Object.run (/opt/homebrew/Cellar/mw/1.7.0/libexec/node_modules/@oclif/core/lib/main.js:97:16)
```

Version info:
```
#> mw --version
 ›   Warning: @mittwald/cli update available from 1.7.0 to 1.9.0.
@mittwald/cli/1.7.0 darwin-arm64 node-v20.19.3
```

LukasFritzeDev:
Have you used `brew` to install the CLI? If so please use `brew upgrade mw` to update.

NeoBlack:
> [@NeoBlack](https://github.com/NeoBlack) That suggestion was implemented in [#1324](https://github.com/mittwald/cli/pull/1324) 😉
That was quick :)

PR #1324, opened 2025-08-28
Add prerun hook for update: Check if installed with brew

LukasFritzeDev:
relates https://github.com/mittwald/cli/issues/1322

### DIFF
--- a/src/hooks/prerun/update-brew-check.ts
+++ b/src/hooks/prerun/update-brew-check.ts
@@ -1,0 +1,24 @@
+import { Hook } from "@oclif/core";
+import { execSync } from "child_process";
+
+const hook: Hook<"prerun"> = async function (opts) {
+  const isInstalledWithBrew = () => {
+    try {
+      const cellar = execSync("brew --cellar", { encoding: "utf8" });
+      return opts.config.root.startsWith(cellar);
+    } catch {
+      return false;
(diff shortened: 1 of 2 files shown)


src/hooks/prerun/update-brew-check.ts
```diff
@@ -0,0 +1,23 @@
...
+    } catch {
+      return false;
+    }
+  };
+
+  if (opts.Command.id === "update") {
+    if (isInstalledWithBrew()) {
+      opts.context.error(
+        "installed with brew.\nUse `brew upgrade mw` to update to the newest version",
+      );
```
martin-helmich, line 18: This will trigger a full-blown error message. Changing to something less intimidating:


```suggestion
      opts.context.log(
        "It looks like the mw CLI was installed using Homebrew.\nPlease use `brew upgrade mw` to update to the newest version",
      );
      opts.context.exit(1);
```

WDYT?

martin-helmich: From the looks of it, I think `.warn` might be a good compromise.

Alternatively, if we use error, we should customize the rendering for that specific error, because the default renderer is more intended for unexpected exceptions (which is why it also suggests the GitHub link and gives you the stack trace), and not for anticipated errors like these.

PR #1569, opened 2026-01-06
Raise node version in docs, mention installation trap

gandie:
Encountered strange error after installing on Ubuntu 24.04. :

```bash
~/Mittwald_Dev $ mw login token
file:///usr/local/lib/node_modules/@mittwald/cli/node_modules/string-width/index.js:16
const zeroWidthClusterRegex = /^(?:\p{Default_Ignorable_Code_Point}|\p{Control}|\p{Mark}|\p{Surrogate})+$/v;
                              ^

SyntaxError: Invalid regular expression flags
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:152:18)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:298:14)

Node.js v18.19.1
```

Solved by simply upgrading node via `nvm` ( avoiding stone age debian package ) , still worth mentioning due to installation traps one can easily run into, especially when ure new to the whole system.

The specific feature causing the error in older JS engines ( it's the `\v` part in RegEx ) :
https://v8.dev/features/regexp-v-flag

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ Features that would (usually) not be added to the CLI:
 
 For a development environment for this CLI, you need:
 
-- Node.JS (>=20.0.7)
+- Node.JS (>=20.7.0)
 - Yarn
 - 7zip + NSIS (for Windows builds)
 - Docker (for building the Docker image)


CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ Features that would (usually) not be added to the CLI:
 
 For a development environment for this CLI, you need:
 
-- Node.JS (>=18)
+- Node.JS (>=20.0.7)
```
martin-helmich: `20.0.7` or `20.7.0`?

PR #1584, opened 2026-01-15
chore: discontinue build of macOS pkg files

martin-helmich:
This PR discontinues the building of macOS pkg files.

The pkg build was a leftover from the past. Moreover, installation via PKG file was never recommended, neither in this project's README, nor in [the developer docs](https://developer.mittwald.de/docs/v2/cli/usage/install/). Furthermore, PKG installation was always cumbersome because we never bothered signing these packages.

For macOS users, the recommended way of installation has always been Homebrew, which will continue to work as usual.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,50 +90,11 @@ jobs:
           name: pkg-windows
           path: dist/win32/*.exe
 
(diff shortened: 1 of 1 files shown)


gandie (approved):
LGTM!

Issue #1323, opened 2025-08-28
"mw stack list" (and others) hang indefinitely on Windows


martin-helmich:
_Converted from customer support ticket_

**Describe the bug**
Some commands like `stack list`, `stack deploy` and others hang indefinitely.

**To Reproduce**
Unclear. Further investigation is required.

**Expected behavior**
Commands should run to completion

**Console logs**
According to debug output, the command hangs immediately _after_ initialization, and before the first API call.

<details>

<summary>Full output with <code>DEBUG=*</code></summary>

```
  oclif:config start init hook +0ms
  oclif:config:ts-path Skipping typescript path lookup for C:\Program Files\mw\client\node_modules\@oclif\plugin-update because NODE_ENV is NOT "test" or "development" +1s
  oclif:config:ts-path Skipping typescript path lookup for C:\Program Files\mw\client\node_modules\@oclif\plugin-warn-if-update-available because NODE_ENV is NOT "test" or "development" +0ms
  oclif:@oclif/plugin-update:hooks:init start (import) C:\Program Files\mw\client\node_modules\@oclif\plugin-update\dist\hooks\init.js +0ms
  oclif:@oclif/plugin-warn-if-update-available:hooks:init start (import) C:\Program Files\mw\client\node_modules\@oclif\plugin-warn-if-update-available\lib\hooks\init\check-update.js +0ms
  oclif:@oclif/plugin-update:hooks:init done +21ms
  oclif:@oclif/plugin-warn-if-update-available:hooks:init done +8ms
  oclif:config init hook done +29ms
  oclif:config runCommand stack:list [] +1ms
  oclif:config:@mittwald/cli (import) C:\Program Files\mw\client\dist\commands\stack\list.js +66ms
  oclif:config start prerun hook +1ms
  oclif:config prerun hook done +1ms
  oclif:mw:stack:list init version: @oclif/core@4.5.2 argv: [] +0ms
  oclif:config start preparse hook +5ms
  oclif:config preparse hook done +0ms
  oclif:config start preparse hook +5ms
  oclif:config preparse hook done +0ms
  oclif:config start preparse hook +1ms
  oclif:config preparse hook done +0ms
```

</details>

**Environment (please complete the following information):**
 - OS: Windows
 - Shell: Powershell
 - Terminal [e.g. iTerm, Warp, ...]
 - Version (output of `mw --version`) [e.g. 1.2.3]

**Additional context**

martin-helmich:
[...]

PR #1774, opened 2026-04-16
fix: flush oclif output buffers on error

gandie:
Trying to narrow down #1323 even further, i've discovered some passages that might also need flushing, as well as one passage in `ExecRenderBaseCommand` that might cause trouble on buffered consoles. Not yet tried on windows machine yet, dev version working fine so far.

Also relaxed cleanup pattern in `package.json` as my cleanups were incomplete with the given one, confusing me when using dev tools mentioned in contribution guide ( `yarn compile` did not rebuild `dist` after `yarn clean` , took some time to figure out root cause: the `tsconfig.tsbuildinfo` file was still present ).

### DIFF
--- a/src/lib/basecommands/RenderBaseCommand.tsx
+++ b/src/lib/basecommands/RenderBaseCommand.tsx
@@ -54,9 +54,12 @@ export abstract class RenderBaseCommand<
 
   public async run(): Promise<void> {
     const onError = () => {
-      setImmediate(() => {
+      setImmediate(async () => {
         handle.unmount();
-        process.exit(1);
+        import("@oclif/core").then(async (oclif) => {
+          await oclif.default.flush();
+          process.exit(1);
+        });
       });
     };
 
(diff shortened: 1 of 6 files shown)


src/lib/basecommands/RenderBaseCommand.tsx
```diff
@@ -54,9 +54,12 @@ export abstract class RenderBaseCommand<
 
   public async run(): Promise<void> {
     const onError = () => {
-      setImmediate(() => {
+      setImmediate(async () => {
         handle.unmount();
-        process.exit(1);
+        import("@oclif/core").then(async (oclif) => {
```
martin-helmich, line 59: Import could be moved to the top of the module, I think.

martin-helmich (changes requested):

gandie:
These changes are red herings as testing showed, closing.

PR #1848, opened 2026-05-19
fix(ci): use relative paths for artifact download target

gandie:
Well. Hard to test. Not sure if local run would have spotted this even.


martin-helmich (approved):
Argh... 🫠

PR #1880, opened 2026-06-02
fix(domain/virtualhost): keep colons in --path-to-url values

### LINKED ISSUE #1879: domain virtualhost create --path-to-url splits URL on every colon, breaking all real URLs
## What happens

Creating an ingress with a URL target fails because the CLI splits the flag value on every `:` instead of only the first one. As a result, the URL after the protocol's `://` is dropped, and the API rejects the truncated value.

**Repro:**

```bash
mw domain virtualhost create --project-id <pid> \
  --hostname www.example.com \
  --path-to-url "/:https://example.com"
```

**Result:**

```
Step 1: creating ingress... FAILED
  Error: AxiosError: Request failed with status code 400
Message  3 INVALID_ARGUMENT: validation failed with
  ['paths.0.Target.Url: Does not match format ''uri''']
```

The request body sent to the API contains `url: "https"` — everything after the second `:` was dropped.

## Root cause

[`src/commands/domain/virtualhost/create.tsx:86`](https://github.com/mittwald/cli/blob/master/src/commands/domain/virtualhost/create.tsx#L86):

```ts
const [path, url] = pathToUrl.split(":");   // splits on every colon
```

For `/:https://example.com` this yields `["/", "https", "//example.com"]`; destructuring keeps only the first two → `url = "https"`.

The sibling `--path-to-container` handler does it correctly with `pathToContainer.split(":", 3)` (and is also safe because its third token contains no further colons in practice).

Note: in JS, `split(":", 2)` would *not* be a correct fix — it drops the rest of the string instead of keeping it. The proper fix is to split on the first colon while preserving the remainder; e.g. `split(":")` followed by destructuring with rest + `join(":")`, or `indexOf + slice`.

## Workaround

Call `POST /v2/ingresses` directly via `curl` / API. The CLI is currently unusable for `--path-to-url`.

## Tested versions

- `@mittwald/cli` 1.13.1 — broken
- `@mittwald/cli` 1.17.1 — broken
- `master` (as of today) — still broken

## Fix

PR opened with a one-line fix.

dkd-dobberkau:
Fixes #1879.

`--path-to-url "/:https://example.com"` currently fails with `Target.Url: Does not match format 'uri'` because the value is split on every colon, so `url` ends up as `"https"`.

This change splits on the first colon while preserving the rest:

```ts
const [path, ...urlParts] = pathToUrl.split(":");
const url = urlParts.join(":");
```

`split(":", 2)` would not work in JS — it drops the rest of the string. The destructure + `join` approach keeps the full URL, including any colons inside it (e.g. ports: `https://backend.example.com:8080/v2`).

## Verification

- `yarn compile`, `yarn lint`, prettier — all green.
- `yarn test:unit` (full suite): 121 tests passing, no regressions.
- Manual check of the parser logic against three realistic inputs:

```
/:https://redirect.example                       → path="/",    url="https://redirect.example"
/api:https://backend.example.com:8080/v2         → path="/api", url="https://backend.example.com:8080/v2"
/old:http://example.com                          → path="/old", url="http://example.com"
```

- Verified end-to-end against the live Mittwald API (POST /v2/ingresses) that the resulting payload is accepted and the 301 redirect works.

No new unit test added — this command does not currently have unit-test coverage and the parsing is inline in `exec()`. Happy to extract a helper and add a test if maintainers prefer.

### DIFF
--- a/src/commands/domain/virtualhost/create.tsx
+++ b/src/commands/domain/virtualhost/create.tsx
@@ -83,7 +83,8 @@ export default class Create extends ExecRenderBaseCommand<
     }
 
     for (const pathToUrl of this.flags["path-to-url"] ?? []) {
-      const [path, url] = pathToUrl.split(":");
+      const [path, ...urlParts] = pathToUrl.split(":");
+      const url = urlParts.join(":");
       paths.push({ path, target: { url } });
     }
 


martin-helmich (approved):
Ouch. 🤦‍♂️  Good catch, thanks. 👍

PR #1954, opened 2026-06-30
feat(app): allow pinning system-software versions on app create

### LINKED ISSUE #1951: `mw app create php|node|python`: allow pinning the system-software (runtime) version at creation
### Summary

`mw app create php` (and the Node.js / Python equivalents) offer no way to choose the runtime / system-software version. The newest available version is assigned by default, and there's no flag to change that at creation time.

### Why this is a problem

For migrations and version-sensitive apps this bites. Example: Shopware 6.6 supports PHP 8.1–8.3, but `mw app create php` provisions the newest available PHP (8.5 at the time of writing), so the app won't start. Today the only path is a second call after creation:

```bash
mw app create php -p <project> --document-root /public
mw app dependency update <installationId> --set php=~8.3
```

The capability exists — `mw app dependency update`, and the API how-to ["Change the version of a system software"](https://developer.mittwald.de/docs/v2/api/howtos/change-systemsoftware-version/) — it just isn't reachable at create time. The documented `POST /v2/projects/{projectId}/app-installations` body only accepts `appVersionId` / `description` / `updatePolicy` (no `systemSoftware`), so the version genuinely can't be pinned in one step today; it's create-then-update.

### Proposed solution

Add a way to pin the system-software version on `mw app create php|node|python`, e.g. a `--set <dependency>=<version>` flag mirroring `mw app dependency update` (or a dedicated `--php-version` / `--node-version`). The command could perform create + dependency-update internally (or pass `systemSoftware` straight through if/when the create endpoint starts accepting it).

### Impact

Removes an easy-to-miss second step. An app provisioned on the wrong major (e.g. PHP 8.5 for a PHP 8.1–8.3 app) fails to start, which is confusing for users who reasonably assume `create` would honour a chosen version.

CLI version: 1.18.0.

martin-helmich:
## Summary

Closes #1951.

`mw app create php|node|python` previously offered no way to choose the runtime / system-software version — the newest available version was always assigned, so version-sensitive apps (e.g. Shopware 6.6 on PHP 8.5) would fail to start and required a second `mw app dependency update` call.

This adds a `--set <dependency>=<version>` flag to the custom-runtime create commands (`php`, `php-worker`, `node`, `python`) that pins the system-software version right after the installation is created, mirroring the existing `mw app dependency update --set` syntax.

```bash
mw app create php --document-root /public --set php=~8.3
mw app create node --set node=~18
```

## Changes

- Extracted the dependency-resolution logic from `app dependency update` into a shared `updateAppDependencies` helper (`src/lib/resources/app/dependencies.tsx`), now used by both the create flow and the update command.
- Added the multi-valued `set` flag to the shared app flag set and wired it into the four custom-runtime create commands via `AppInstaller`.
- `AppInstaller.exec` applies the pinned versions (update policy `patchLevel`) after creating the installation and before waiting for it to normalize.
- Filtered non-scalar flags out of the `userInputs` payload in `triggerAppInstallation` so the array-valued `set` flag isn't serialized as a user input.
- Added usage examples to the `php`, `node`, and `python` create commands.

> Note: the README/docs could not be regenerated in this environment (`oclif readme` fails with an unrelated `LRUCache is not a constructor` error); docs should be regenerated before merge.


martin-helmich:
Converted back to draft status; there's an API change in the works that would allow us to set the system software versions with a single API call while installing the app.

src/lib/resources/app/dependencies.tsx
```diff
@@ -0,0 +1,212 @@
...
+ * Resolves a list of `<dependency>=<version>` specifications into a map of
+ * system-software id to the concrete version it should be pinned to.
+ *
+ * `<dependency>` is the name of a system software (e.g. `php`) and `<version>`
+ * is a semver constraint (e.g. `~8.3`). For each spec, the matching available
+ * version is resolved.
+ */
+export async function resolveSystemSoftwareUpdates(
+  apiClient: MittwaldAPIV2Client,
+  process: ProcessRenderer,
```
gandie, line 31: Not a blocker, but maybe worth a note: Passing the `ProcessRenderer` down ( to `resolveSystemSoftwareVersion` ) ties these sequences of API calls closely to CLI. This is fine and does not surprise as we're working on CLI here. Nevertheless i would prefer to keep the `ProcessRenderer` in the command, acting as orchestration layer. This would allow these functions to work without process rendering knowledge.

gandie (approved):

PR #1973, opened 2026-07-02
fix(app): surface uninstall failures instead of reporting false success

### LINKED ISSUE #1970: `mw app uninstall` reports success despite failing with HTTP 412 (linked primary DB)
### Summary

`mw app uninstall <app> --force` prints **both** an error **and** "Process completed successfully" when the app has a linked **primary** database, and the app is **not** uninstalled. An automated / agent-driven run can't tell that it failed.

Verified against **`mw 1.19.0`**.

### Repro

1. An app installation with a linked primary database (`PATCH /v2/app-installations/{id}/database`, `purpose: primary`).
2. `mw app uninstall <shortId> --force`

Output:

```
  Error: AxiosError: Request failed with status code 412
Completed: Process completed successfully

Summary: [Complex content]
```

Afterwards the app is still present (`mw app list`) — the uninstall actually **failed** (HTTP 412), yet the CLI reports success (and appears to exit 0).

### Two problems

1. **False success / wrong exit semantics.** On a 412 the CLI should surface the failure, not print "Process completed successfully", and exit non-zero. Today a script cannot detect the failure.
2. **The 412 reason isn't surfaced.** The cause is a linked **primary** database. A primary DB also can't be unlinked directly — `DELETE /v2/app-installations/{id}/databases/{databaseId}` returns `412 "primary database of app installation cannot be unlinked"`; you must first repurpose it to `custom`/`cache`, then unlink. A clear message like *"cannot uninstall: app has a linked primary database — unlink or repurpose it first"* would save a lot of guessing.

### Note — why the guard matters

The raw API `DELETE /v2/app-installations/{id}` **does** cascade-delete the linked primary database. The CLI's 412 guard is a good safety net against silent data loss — it just needs to *report* the refusal correctly instead of claiming success.

### Context

Surfaced while validating the [`mittwald-migrate`](https://github.com/mittwald/agent-skills) skill.

martin-helmich:
## Problem

`mw app uninstall <id> --force` printed **both** an error and `Process completed successfully`, and appeared to exit `0`, even when the uninstall actually failed with HTTP `412` (the app has a linked **primary** database). Scripts could not detect the failure, and the reason was not surfaced.

## Root cause

`DeleteBaseCommand.exec()` caught errors thrown by `deleteResource()` and marked the step as failed, but then still called `process.complete(...)`. The `SimpleProcessRenderer` (used in non-TTY/CI contexts) unconditionally writes `Completed: Process completed successfully` in `complete()`, so a failed deletion was reported as a success.

## Fix

- `DeleteBaseCommand` now calls `process.error(err)` in the catch branch instead of `process.complete(...)`, so the failure is surfaced. It keeps `ux.exit(1)` to guarantee a non-zero exit code. This corrects **all** delete commands.
- `app uninstall` now maps the HTTP `412` (linked primary database) to an actionable message — "cannot uninstall: app has a linked primary database — unlink or repurpose it first ..." — while passing every other error through unchanged (with the original error attached as `cause`).

## Tests

- `src/lib/basecommands/DeleteBaseCommand.test.tsx` — asserts that on a failing deletion the command does **not** report success, surfaces the error via `process.error`, and exits non-zero (`ExitError` with `oclif.exit === 1`); and that a successful deletion still reports success and does not exit non-zero.
- `src/commands/app/uninstall.test.ts` — covers the `412` → primary-database message mapping and pass-through of other errors.

Closes #1970

### DIFF
--- a/src/commands/app/uninstall.ts
+++ b/src/commands/app/uninstall.ts
@@ -1,17 +1,63 @@
 import { assertStatus } from "@mittwald/api-client-commons";
-import axios from "axios";
 import { DeleteBaseCommand } from "../../lib/basecommands/DeleteBaseCommand.js";
 import { appInstallationArgs } from "../../lib/resources/app/flags.js";
+
+interface APIErrorBody {
+  message?: string;
+  type?: string;
+}
+
+interface APIErrorLike {
+  response?: {
+    data?: unknown;
+    status?: number;
+  };
+}
+
+function getAPIErrorDetails(err: unknown): {
+  body: APIErrorBody | undefined;
+  status: number;
+} | null {
+  if (typeof err !== "object" || err === null) {
+    return null;
+  }
+
+  const response = (err as APIErrorLike).response;
+  if (typeof response?.status !== "number") {
+    return null;
+  }
+
+  const body =
+    typeof response.data === "object" && response.data !== null
+      ? (response.data as APIErrorBody)
+      : undefined;
+
+  return {
+    status: response.status,
+    body,
+  };
+}
+
+function isPrimaryDatabasePreconditionError(
+  body: APIErrorBody | undefined,
+): boolean {
+  const signature = `${body?.type ?? ""} ${body?.message ?? ""}`.toLowerCase();
+  return signature.includes("primary") && signature.includes("database");
+}
 
 /**
  * Maps an error raised while uninstalling an app installation to a more
  * actionable one. The API answers with an HTTP 412 (Precondition Failed) when
- * the app still has a linked _primary_ database; in that case we surface a
- * clear hint instead of a raw Axios error. All other errors are passed through
- * unchanged so they are not accidentally swallowed.
+ * the app still has a linked _primary_ database. We therefore require both the
+ * precondition status and a matching error signature before rewriting the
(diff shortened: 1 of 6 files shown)


gandie (approved):

src/commands/app/uninstall.ts
```diff
@@ -1,7 +1,28 @@
...
 
+/**
+ * Maps an error raised while uninstalling an app installation to a more
+ * actionable one. The API answers with an HTTP 412 (Precondition Failed) when
+ * the app still has a linked _primary_ database; in that case we surface a
+ * clear hint instead of a raw Axios error. All other errors are passed through
+ * unchanged so they are not accidentally swallowed.
+ */
+export function mapUninstallError(err: unknown): unknown {
+  if (axios.isAxiosError(err) && err.response?.status === 412) {
```
martin-helmich: Unsure if this can't be done more elegantly. Also don't know if there are other conditions in which a 412 status code might arise. Should investigate if there's a more specific error code in the response body that we could filter for.

gandie: @martin-helmich refactor is in place, have a look

PR #2005, opened 2026-07-15
feat(database): add "database mysql upgrade" command

martin-helmich:
Adds a `database mysql upgrade` command for upgrading a MySQL database to a newer version.

## Background

The API has no dedicated upgrade operation. `PATCH /v2/mysql-databases/{id}` accepts an optional `version` field alongside `description` and `characterSettings`, and that is what this command drives.

## Behaviour

- Upgrade candidates come from `GET /v2/mysql-versions`, filtered to enabled versions newer than the database's current one. MySQL does not support downgrades, so older versions are never offered.
- Versions are compared numerically via semver coercion. The `<major>.<minor>` format is not valid semver, and a lexical sort would place `10.11` before `8.0`.
- `--version` takes an explicit version or `latest`; if omitted, the version is prompted for interactively, following `app upgrade`.
- If the database already runs the newest available version, the command reports that and exits without patching.
- An invalid `--version` fails with the list of valid targets rather than sending a doomed request.
- Upgrades are confirmed unless `--force` is passed, since they are disruptive.
- `--wait` polls until the upgrade has actually completed (see below).

## Notes for reviewers

The interesting part is `--wait`. Two API behaviours make the obvious implementation wrong, both found by testing against a real database:

- `version` holds the **desired** version and is updated as soon as the patch is accepted, so it does not indicate that the upgrade finished.
- `status` still reads `ready` from before the upgrade for a moment after patching.

Together these mean a naive "target version and `ready`" check returns before the upgrade even starts, leaving the database at `pending`. Unlike `AppAppInstallation`, `DatabaseMySqlDatabase` has no current/desired pair, so the `current == desired` approach used by `waitUntilAppStateHasNormalized` is not available here.

Instead, `statusSetAt` is captured before patching and required to have changed before `ready` is trusted, proving the database changed state at least once since. Both timestamps come from the API, so the comparison is immune to client clock skew. This is `isUpgradeComplete()` in `lib/resources/database/mysql/upgrade.ts`, covered by unit tests including the stale-`ready` regression.

The existing command tests in this area are all `it.skip`'d ("to be fixed later"), so rather than add another skipped test, the version-selection and completion logic are extracted into `lib/resources/database/mysql/` and covered by real unit tests.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -14,6 +14,7 @@ Manage databases (like MySQL and Redis) in your projects
 * [`mw database mysql phpmyadmin DATABASE-ID`](#mw-database-mysql-phpmyadmin-database-id)
 * [`mw database mysql port-forward DATABASE-ID`](#mw-database-mysql-port-forward-database-id)
 * [`mw database mysql shell DATABASE-ID`](#mw-database-mysql-shell-database-id)
+* [`mw database mysql upgrade DATABASE-ID`](#mw-database-mysql-upgrade-database-id)
 * [`mw database mysql user create`](#mw-database-mysql-user-create)
 * [`mw database mysql user delete USER-ID`](#mw-database-mysql-user-delete-user-id)
 * [`mw database mysql user get ID`](#mw-database-mysql-user-get-id)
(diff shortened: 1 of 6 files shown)


martin-helmich:
Verified via manual testing ✅

PR #2023, opened 2026-07-21
feat(container): support env var expansion in consuming commands

### LINKED ISSUE #1572: "docker run" compatibility for environment variables without values
After #1571, `mw container run` will raise an Error when just passing the name of an environment variable as flag (like `--env FOO`). Before, this was just undefined behaviour, in which the environment variable would (probably) have been set to an empty null value.

However, `docker run` actually behaves differently in this case, and would extract the value of the environment variable from the users local environment, instead.

> Careful; this actually works with the regular old `docker run` command; in which case (`--env FOO`), the value of the environment variable is actually just read from your shell environment. 

 _Originally posted by @martin-helmich in [#1571](https://github.com/mittwald/cli/pull/1571/changes#r2668477908)_

gandie:
Resolve #1572 

Tested in fish shell:

```fish
set -x POSTGRES_PASSWORD 'yoursecrethere'
mw container run -e POSTGRES_PASSWORD postgres:latest
```

Container spawns with password and starts without errors.

### DIFF
--- a/src/commands/container/run.tsx
+++ b/src/commands/container/run.tsx
@@ -46,7 +46,7 @@ export class Run extends ExecRenderBaseCommand<typeof Run, Result> {
     env: Flags.string({
       summary: "set environment variables in the container",
       description:
-        "Format: KEY=VALUE. Multiple environment variables can be specified with multiple --env flags.",
+        "Format: KEY=VALUE or KEY. If only KEY is provided, the value is resolved from the caller environment (exported variables only). Multiple environment variables can be specified with multiple --env flags.",
       required: false,
       multiple: true,
       multipleNonGreedy: true,
--- a/docs/container.md
+++ b/docs/container.md
@@ -551,7 +551,8 @@ FLAG DESCRIPTIONS
 
   -e, --env=<value>...  set environment variables in the container
 
-    Format: KEY=VALUE. Multiple environment variables can be specified with multiple --env flags.
+    Format: KEY=VALUE or KEY. If only KEY is provided, the value is resolved from the caller environment (exported
+    variables only). Multiple environment variables can be specified with multiple --env flags.
 
   -m, --memory=<value>  set memory limit for the container
 
@@ -789,7 +790,8 @@ FLAG DESCRIPTIONS
 
   -e, --env=<value>...  set environment variables in the container
 
-    Format: KEY=VALUE. Multiple environment variables can be specified with multiple --env flags.
+    Format: KEY=VALUE or KEY. If only KEY is provided, the value is resolved from the caller environment (exported
+    variables only). Multiple environment variables can be specified with multiple --env flags.
 
   -p, --project-id=<value>  ID or short ID of a project; this flag is optional if a default project is set in the context
 
--- a/docs/experimental.md
+++ b/docs/experimental.md
@@ -58,7 +58,8 @@ EXAMPLES
 FLAG DESCRIPTIONS
   -e, --env=<value>...  set environment variables in the container
 
-    Format: KEY=VALUE. Multiple environment variables can be specified with multiple --env flags.
+    Format: KEY=VALUE or KEY. If only KEY is provided, the value is resolved from the caller environment (exported
+    variables only). Multiple environment variables can be specified with multiple --env flags.
 
   -p, --project-id=<value>  ID or short ID of a project; this flag is optional if a default project is set in the context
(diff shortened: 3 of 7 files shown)


src/commands/container/run.tsx
martin-helmich: I'm seeing this exact flag definition repeated three times.
This is not a new issue, but maybe now's the time to extract all these container flag definitions into a common file?

martin-helmich (approved):

src/commands/container/run.tsx
gandie: Done. Some flags have slightly different descriptions for adding and updating, e.g. `description`. That's why flag factories allow arguments for those. Maybe this is slightly overengineered now, maybe it is just prepared for future command additions.

src/lib/resources/container/common-flags.ts
```diff
@@ -0,0 +1,106 @@
...
+type ContainerPublishFlagOptions = ContainerStringFlagOptions & {
+  char?: AlphabetLowercase;
+  multipleNonGreedy?: boolean;
+};
+
+type ContainerVolumeFlagOptions = ContainerStringFlagOptions & {
+  multipleNonGreedy?: boolean;
+};
+
+export function makeContainerEnvFlag(options: MultiValueFlagOptions = {}) {
```
martin-helmich: Note that other commands typically use the `Flags.custom` feature to build configurable flag. See `backupScheduleFlagDefinitions` for reference (as example).

src/lib/resources/container/common-flags.ts
```diff
@@ -0,0 +1,106 @@
...
+  return Flags.string({
+    summary: "set environment variables in the container",
+    description:
+      "Format: KEY=VALUE or KEY. If only KEY is provided, the value is resolved from the caller environment (exported variables only). Multiple environment variables can be specified with multiple --env flags.",
+    required: false,
+    multiple: true,
+    multipleNonGreedy: options.multipleNonGreedy,
```
martin-helmich: I'm suspecting this should _always_ be `true` (it was false on `update`, but that may have been an oversight).

src/lib/resources/container/common-flags.ts
```diff
@@ -0,0 +1,95 @@
...
+  "If you specify a file path as volume, this will mount a path from your hosting environment's file system (NOT your local file system) into the container. " +
+  "You can also specify a named volume, which needs to be created beforehand.";
+
+export function makeContainerDescriptionFlagOptions(summary: string) {
+  return Flags.custom<string>({
+    summary,
+    description: "This helps identify the container's purpose or contents.",
+    required: false,
+  })();
+}
```
martin-helmich: `Flags.custom` already exports a constructor function; the idiomatic usage would be this:

```
// Export
export const containerDescriptionFlag = Flags.custom<string>({
  summary,
  description: "This helps identify the container's purpose or contents.",
  required: false,
});

// Usage
const flags = {
  description = containerDescriptionFlag({
    summary: "boo"
  })
}
```

src/commands/container/run.tsx
```diff
@@ -43,72 +53,37 @@ export class Run extends ExecRenderBaseCommand<typeof Run, Result> {
...
+    description: containerDescriptionFlag({
       summary: "add a descriptive label to the container",
-      description: "This helps identify the container's purpose or contents.",
-      required: false,
     }),
-    entrypoint: Flags.string({
+    entrypoint: containerEntrypointFlag({
       summary: "override the default entrypoint of the container image",
       description:
         "The entrypoint is the command that will be executed when the container starts. If omitted, the entrypoint defined in the image will be used.",
```
martin-helmich: I think we could also move summary+description to the custom flag definition

src/commands/container/run.tsx
```diff
@@ -43,72 +53,37 @@ export class Run extends ExecRenderBaseCommand<typeof Run, Result> {
...
     name: Flags.string({
       summary: "assign a custom name to the container",
       description:
         "This makes it easier to reference the container in subsequent commands. If omitted, a random name will be generated automatically.",
       required: false,
     }),
-    publish: Flags.string({
+    publish: containerPublishFlag({
       summary: "publish a container's port(s)",
       description:
```
martin-helmich: I think we could also move summary+description to the custom flag definition

martin-helmich (changes requested):

PR #2033, opened 2026-07-23
feat(context): enrich context getter to provide project details

### LINKED ISSUE #929: mw context show would be very useful
**Is your feature request related to a problem? Please describe.**
I see context get and set.
But if a project is lying for some time, i may forget which one is the remote context.

**Describe the solution you'd like**
With `mw context show` i'll get an clue about it, if it would show up the "connected" database, project and app.

**Describe alternatives you've considered**
-
**Additional context**
Context is the mw ddev environment whoch was created by mw ddev pull

gandie:
Resolves #929

### DIFF
--- a/src/commands/context/get.tsx
+++ b/src/commands/context/get.tsx
@@ -5,15 +5,62 @@ import { Value } from "../../rendering/react/components/Value.js";
 import { usePromise } from "@mittwald/react-use-promise";
 import { Note } from "../../rendering/react/components/Note.js";
 import { Box, Text } from "ink";
-import { Set } from "./set.js";
+import { Set as SetCommand } from "./set.js";
 import { RenderJson } from "../../rendering/react/json/RenderJson.js";
 import { useRenderContext } from "../../rendering/react/context.js";
 import { LocalFilename } from "../../rendering/react/components/LocalFilename.js";
+import { MittwaldAPIV2 } from "@mittwald/api-client";
+import { assertStatus } from "@mittwald/api-client-commons";
 import Context, {
   ContextKey,
   ContextValue,
   ContextValueSource,
 } from "../../lib/context/Context.js";
+import {
+  getAppFromUuid,
+  getAppInstallationFromUuid,
+} from "../../lib/resources/app/uuid.js";
+
+type AppLinkedDatabase = MittwaldAPIV2.Components.Schemas.AppLinkedDatabase;
+
+type LinkedDatabaseSummary = {
+  databaseId: string;
+  purpose: string;
+  kind: "mysql" | "redis" | "unknown";
+  name?: string;
+};
+
+type AppSummary = {
+  installationId: string;
+  appId: string;
+  appName: string;
+  installationPath: string;
+  linkedDatabases: LinkedDatabaseSummary[];
+};
+
+type StackSummary = {
+  id: string;
+  description?: string;
+  services: number;
+  volumes: number;
+};
+
+type ContainerSummary = {
+  id: string;
+  name: string;
+  status: string;
+  stackId?: string;
+};
+
+type ProjectOverview = {
+  projectId?: string;
+  projectName?: string;
+  resolvedFrom?: "project-id" | "installation-id";
+  apps: AppSummary[];
+  stacks: StackSummary[];
+  containers: ContainerSummary[];
+  unavailableReason?: string;
+};
 
 const ContextSourceValue: FC<{ source: ContextValueSource }> = ({ source }) => {
   switch (source.type) {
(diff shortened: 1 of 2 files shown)


src/commands/context/get.tsx
```diff
@@ -75,9 +122,90 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
...
+                </Text>
+              ))
+            ) : (
+              <Text color="gray">no linked databases</Text>
+            )}
+          </Box>
+        ))}
+      </Box>
+    );
+  } else {
```
martin-helmich: Do I read this correctly that this shows _either_ app installation or container stacks? Mind that a project might contain _both_ 😉

src/commands/context/get.tsx
```diff
@@ -75,9 +122,90 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
...
+    "Resolved from": <Value>{overview.resolvedFrom ?? "project-id"}</Value>,
+  };
+
+  if (overview.apps.length > 0) {
+    rows["Apps"] = (
+      <Box flexDirection="column">
+        {overview.apps.map((app) => (
+          <Box key={app.installationId} flexDirection="column" marginBottom={1}>
+            <Text>
+              <Value>{app.appName}</Value>{" "}
```
martin-helmich: 💡 Idea: If a specific app installation is set in the context, highlight that particular app installation in the list?

src/commands/context/get.tsx
```diff
@@ -75,9 +122,90 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
...
+        Project overview is unavailable: {overview.unavailableReason}
+      </Note>
+    );
+  }
+
+  const rows: Record<string, ReactNode> = {
+    Project: (
+      <Text>
+        <Value>{overview.projectName ?? overview.projectId}</Value>{" "}
+        <Text color="gray">({overview.projectId})</Text>
```
martin-helmich: Show short ID instead of full ID

src/commands/context/get.tsx
```diff
@@ -75,9 +122,90 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
...
+      </Text>
+    ),
+    "Resolved from": <Value>{overview.resolvedFrom ?? "project-id"}</Value>,
```
martin-helmich: Maybe condense `Project` and `Resolved from` into a single row? 🤔

src/commands/context/get.tsx
```diff
@@ -109,15 +237,221 @@ const GetContext: FC<{ ctx: Context }> = ({ ctx }) => {
...
+      } catch {
+        return {
+          unavailableReason: "could not resolve project from installation-id",
+        };
+      }
+    },
+    [projectIdFromContext, appInstallationId],
+  );
+
+  const overview = usePromise(
```
martin-helmich: This entire part is uncomfortably long. Consider refactoring it, also maybe splitting this file into smaller modules.

martin-helmich (changes requested):

src/lib/context/projectOverview.ts
```diff
@@ -0,0 +1,251 @@
...
+    });
+
+    let stacks: StackSummary[] = [];
+    let containers: ContainerSummary[] = [];
+
+    try {
+      const [stackResponse, serviceResponse] = await Promise.all([
+        apiClient.container.listStacks({ projectId }),
+        apiClient.container.listServices({ projectId }),
+      ]);
```
martin-helmich: This seems a bit redundant; the stack list should already contain the services (grouped by stack).

src/commands/context/get.tsx
```diff
@@ -75,9 +81,122 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
...
+    );
+
+  rows["Stacks"] =
+    overview.stacks.length > 0 ? (
+      <Box flexDirection="column">
+        <Text>
+          <Value>{overview.stacks.length}</Value> total
+        </Text>
+        {overview.stacks.slice(0, 5).map((stack) => (
+          <Text key={stack.id} color="gray">
```
martin-helmich: Formatting: Why are apps all formatted as `"green"`, and stacks all as `"gray"`? Seems a bit inconsistent.

Idea: Format the specific app/stack set in the context as green, all other gray?

src/commands/context/get.tsx
```diff
@@ -75,9 +81,122 @@ const ContextSource: FC<{ source: ContextValueSource }> = ({ source }) => {
   );
 };
 
+const ProjectOverviewSection: FC<{ overview: ProjectOverview }> = ({
```
martin-helmich: This component is very long. Consider splitting it into separate subcomponents.

src/lib/context/projectOverview.ts
```diff
@@ -0,0 +1,251 @@
...
+      const redisResponse = await apiClient.database.listRedisDatabases({
+        projectId,
+      });
+      assertStatus(redisResponse, 200);
+      for (const db of redisResponse.data) {
+        databaseById.set(db.id, { name: db.name, kind: "redis" });
+      }
+    } catch {
+      // best effort
+    }
```
martin-helmich: This function contains many chained `await` calls, some of which might be run concurrently using `Promise.all`.

src/commands/context/get.tsx
gandie: Nice one. Giving proper meaning to colors enhanced readability a lot, as you can now now tell what's in context directly via green text, while derived project information text is gray. This allows to see what's affected by context-driven commands from a glance :+1:

src/lib/context/projectOverview.ts
```diff
@@ -0,0 +1,298 @@
...
+  id: string;
+};
+
+export function formatOverviewEntry({
+  shortId,
+  name,
+  status,
+  id,
+}: OverviewEntryData): string {
+  return `${shortId ?? ""} ( ${name} ): ${status} ( ${id} )`;
```
martin-helmich: This looks weird for resources that do not have a short ID.

Maybe something like this?

```suggestion
  if (shortId) {
    return `${name} (${shortId}): ${status} (${id})`;
  } else {
    return `${name}: ${status} (${id})`;
  }
```

martin-helmich (approved):

